PR #49, opened 2015-01-23
Fixed TKIP decryption. Now packets from AP to STA are also decrypted.

joerango:


mfontanini:
Thanks for this PR. Would you mind if I used the pcap file you sent to the forum as a test for this?

joerango:
Sure. You can use the file.

Issue #50, opened 2015-02-05
Preserving timestamps in FileSniffer sniff_loop


EricMCornelius:
First off, wanted to say thank you for all the hard work on this library. Certainly beats rolling all this myself.

I'm wondering if there's any means to retain the pcap header timestamps through the sniff_loop execution. It appears (though please correct me if I'm mistaken) that there is no overload to take in a loop functor with a Packet& argument instead of a PDU, and as near as I can tell that effectively makes retrieving the original pcap pkthdr timestamp impossible via the loop api.

There are a few ways I could see solving this with either new loop methods, overloads, or some relatively gnarly SFINAE, but I'm not sure what you'd have in mind.

PR #51, opened 2015-02-05
Use type_traits to enable Packet& sniff_loop callback variant in C++11

EricMCornelius:


EricMCornelius:
See #50 - this is the beginnings of a potential solution, though I'm not terribly comfortable with the iterator return type change without knowing the code base better.  I'm not familiar with cmake and had no luck getting the tests to build after a brief perusal of: http://www.vtk.org/Wiki/CMake/Testing_With_CTest

include/tins/sniffer.h
```diff
@@ -573,13 +577,44 @@ namespace Tins {
         unsigned _timeout;
     };
 
+    #if TINS_IS_CXX11
+
+    template<bool B, class T = void>
```
mfontanini: This looks good. Could you move this structs/helper functions to internals.h, inside the Internals namespace? It's the only thing that I'd change.

EricMCornelius:
How's that look?  I've never actually seen a callable trait like this before, but I don't see why it wouldn't generalize to variadic packs... should see whether there's a TR out there for it.

include/tins/sniffer.h
```diff
@@ -578,8 +578,13 @@ namespace Tins {
         for(iterator it = begin(); it != end(); ++it) {
             try {
                 // If the functor returns false, we're done
+                #if TINS_IS_CXX11
+                if (!Tins::Internals::invoke_loop_cb<Functor, Packet>(function, *it))
```
mfontanini: I don't see a reason to invoke the function using explicit template parameters. Is there some case in which template parameter deduction doesn't work as expected here?

include/tins/internals.h
```diff
@@ -200,6 +204,34 @@ template<>
...
+template<bool B, class T = void>
+using enable_if_t = typename std::enable_if<B,T>::type;
+
+// Template metaprogramming trait to determine if a functor can accept another parameter as an argument
+template <class T, class P, class=void>
+struct accepts_type : std::false_type { };
+
+template <class T, class P>
+struct accepts_type<T, P, enable_if_t<
+  std::is_same<  decltype(  std::declval<T>()(std::declval<P>())  ), bool>::value
```
mfontanini: This trait seems to not work if your functor has this signature ```bool(Packet&)```, which is one of the compatible signatures for the functor when taking PDUs, so I think it should be extended to Packets as well. This can be fixed by using ```declval<P&>```, which seems very specific for this use, but so is this trait.

include/tins/internals.h
```diff
@@ -200,6 +204,34 @@ template<>
...
+>> : std::true_type { };
+
+// use enable_if to invoke the Packet& version of the sniff_loop handler if possible - otherwise fail to old behavior
+template <class Functor, class Packet>
+bool invoke_loop_cb(Functor& f, Packet& p, typename std::enable_if<accepts_type<Functor, Packet>::value, bool>::type* = 0) {
+  return f(p);
```
mfontanini: Using ```f(std::move(p))``` would be great here. This can't be done with the PDU overload, but being able to take Packets by value on the callback would be really good.

mfontanini:
This looks really good! Using Packets on the callback was something that I've always wanted to add.

include/tins/internals.h
EricMCornelius: If I use a move/forward approach, a non-rvalue reference won't be a valid signature.   Are you good with just supporting:

    bool handler(const Packet&);
    bool handler(Packet&&);
    bool handler(Packet);

?

mfontanini: Mmm, no, I guess it's better if it's compatible with the other signatures, so just taking const and non-const references would be fine.

mfontanini:
Sorry I didn't answer in a few days. I think it was better before, without the move call. That way it's more consistent, since you'll have these valid signatures:

```
bool (const PDU&);
bool (PDU&);
bool (const Packet&);
bool (Packet&);
bool (Packet); // < this one copies, but that's the user's problem, as long as it's documented.
```

Otherwise, it just would be strange if you could have a functor taking a ```PDU&``` but not a ```Packet&```. And still, you can use move semantics on the parameter when taking a ```Packet&```.

EricMCornelius:
Sorry for the delay.  This extends the SFINAE check to enable:

    bool handler(PDU& pdu);
    bool handler(const PDU& pdu);
    bool handler(Packet pkt);
    bool handler(Packet& pkt);
    bool handler(const Packet& pkt);

mfontanini:
Sorry it took me a few days to answer. There seems to be a problem now when you compile on non-c++11 mode. 

When using a callback that takes a PDU const and non-const reference, the compilation error happens on sniffer.h, line 588:

```
if(!function(*it))
```

I guess this should be ```function(it->pdu())```, since the iterator's value_type is now a ```Packet```s rather than ```PDU```s. 

Note that this change will break any code that uses a SnifferIterator. I would normally be against this, but I agree that you should get Packets rather than PDUs from the iterators and I don't really think anyone uses iterators to iterate over a Sniffer anyway.

I'm sorry this is taking a while, but I don't really want to merge this until I'm sure all the combinations work (taking PDUs, Packets, using C++11/03, etc).

EricMCornelius:
Tested locally with c++11 disabled after this change (been ages since I did that last :) )

mfontanini:
Thank you very much! I'll update the documentation/tutorial on the weekend. This is a very good enhancement!

PR #54, opened 2015-02-10
Fixed big endian version for flags_type

UlfWetzker:
Byte 1 and 3 in the big endian version of flags_type seems to be mixed up.
Unfortunately I can't test it.

bit number | litte endian | big endian                                          
--------------|----------------|--------------                                          
0 | tsft:1, | lock_quality:1,                                                    
1 | flags:1, | dbm_noise:1,                                                     
2 | rate:1, | dbm_signal:1,                                                     
3 | channel:1, | fhss:1,                                                        
4 | fhss:1, | channel:1,                                                        
5 | dbm_signal:1, | rate:1,                                                     
6 | dbm_noise:1, | flags:1,                                                     
7 | lock_quality:1, | tsft:1,                                                   
 | |                                                                            
8 | tx_attenuation:1, | reserved3:1,                                            
9 | db_tx_attenuation:1, | rx_flags:1,                                          
10 | dbm_tx_power:1, | db_tx_attenuation:1,                                     
11 | antenna:1, | dbm_tx_power:1,                                               
12 | db_signal:1, | antenna:1,                                                  
13 | db_noise:1, | db_signal:1,                                                 
14 | rx_flags:1, | db_noise:1,                                                  
15 | reserved1:3, | tx_attenuation:1,                                           
 | |                                                                            
16 | | reserved2:5,                                                             
17 | |                                                                          
18 | channel_plus:1, |                                                          
19 | reserved2:12, |                                                            
20 | |                                                                          
21 | | channel_plus:1,                                                          
22 | | reserved1:2,                                                             
23 | |                                                                          
 | |                                                                        
24 | | reserved4:7,                                                             
25 | |                                                                          
26 | |                                                                          
27 | |                                                                          
28 | |                                                                          
29 | |                                                                          
30 | |                                                                          
31 | ext:1; | ext:1;

### DIFF
--- a/include/tins/radiotap.h
+++ b/include/tins/radiotap.h
@@ -390,17 +390,17 @@ namespace Tins {
                     tsft:1,
                     reserved3:1,
                     rx_flags:1,
-                    db_tx_attenuation:1,
-                    dbm_tx_power:1,
-                    antenna:1,
-                    db_signal:1,
                     db_noise:1,
+                    db_signal:1,
+                    antenna:1,
+                    dbm_tx_power:1,
+                    db_tx_attenuation:1,
                     tx_attenuation:1,
                     reserved2:5,
                     channel_plus:1,
                     reserved1:2,
-                    reserved4:7,
-                    ext:1;
+                    ext:1,
+                    reserved4:7;
             } TINS_END_PACK;
         #endif
 


UlfWetzker:
Today I had some doubts about this fix since I read the endianness.h and saw that it only deals with byte-endianness. While reading radiotap.h I always thought that the big and little endian version of flags_type has to handle byte-endianness and bit-endianness caused by serialization to transmit frames over ethernet. So now I am not sure if this fix is correct or not. The endianness of Byte 1 and 3 still looks suspicious to me.

mfontanini:
I'll try to have a look at this tonight. Determining the right position of the fields is tricky. I don't know if there are tests for the fields you modified where the input comes from a real packet (normally the tests are called _ConstructorFromBuffer_) , but if there were, then there are likely to be well positioned.

mfontanini:
For some reason I'm having trouble running the tests after cross compiling for a big endian MIPS architecture. 

However, having a look at the tests, I see that there are some tests that construct a _Radiotap_ from a buffer which check for the _antenna_ field. All of those tests ran successfully on that MIPS system after the last changes I applied to RadioTap, so at least that field is for sure okay.

Do you have any evidence that any of those fields are actually in the wrong order, or is it that by looking at the order, you think they might be wrong? I agree that they look strange though. I basically wrote the big endian fields order by trial and error, since it's pretty tricky to get them right given that the protocol's definition is given in little endian.

UlfWetzker:
I don't have any evidence that these fields are wrong. I just stumbled across these strange looking byte orders.
Maybe I will try to build a big-endian open-wrt on a MIPS based wifi-router during the next weeks to get a look at a real system. I will get back to you when I have a deeper insight on this.

mfontanini:
Great, thank you very much!

mfontanini:
Closing this since there is no real evidence that shows that the ordering of the fields are wrong.

Issue #55, opened 2015-03-03
Radiotap parsing fails on Arduino Yun (OpenWrt)


haralduna:
I am using libtins successfully for live packet capture under OpenWrt. However, parsing of the radiotap header fails, only giving garbage. I have collected a pcap file taken with tcpdump from the Yun. This file shows valid header information in Wireshark, but the header is not correctly read by libtins (on my Ubuntu machine). This leads me to believe that the radiotap header from the Yun (Atheros AR9331) is not read correctly by libtins.

haralduna:
I can provide the pcap dump file on request.

haralduna:
Code failing on Yun:
bool BeaconSniffer::callback(PDU &pdu) {
        // Get the Dot11 layer
        const RadioTap *rtap = pdu.find_pdu<RadioTap>();        
        std::cout
                << std::hex << std::uppercase
//              << "|pl:" << (uint32_t)rtap->channel_plus()
                << "|tp:" << (uint16_t)rtap->channel_type()
                << std::dec
                << "|fq:" << (uint16_t)rtap->channel_freq()
                << "|rt:" << (uint16_t)rtap->rate()
                << "|s1:" << 256-(uint8_t)rtap->dbm_signal();
//              << "|s2:" << (uint8_t)rtap->db_signal()
//              << "|ns:" << (uint8_t)rtap->dbm_noise()
//              << "|sq:" << (uint16_t)rtap->signal_quality();

mfontanini:
Could you please attach the pcap file?

Thanks!

haralduna:
Den 03. mars 2015 16:37, skrev Matias Fontanini:
> Could you please attach the pcap file?
>
> Thanks!
> On Mar 3, 2015 5:07 AM, "haralduna" <notifications@github.com> wrote:
>
> > I can provide the pcap dump file on request.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/mfontanini/libtins/issues/55#issuecomment-76941829>.
> >
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/mfontanini/libtins/issues/55#issuecomment-76968584>.
>
It was already added, check my updated comment

haralduna:
I did some more experiments on the target, with this code:

bool Capture::onePacket(PDU &pdu) {
    const RadioTap *rtap = pdu.find_pdu<RadioTap>();
    if (rtap->present() & RadioTap::PresentFlags::FHSS)
[...]

PR #56, opened 2015-03-04
Fixed radiotap parsing #55

haralduna:
I got it working with the below changes, both on the Yun, and for reading of the pcap file from Yun on a Ubuntu machine.

The header file was clearly wrong in the ordering of the fields for the big endian version. I also had to align the content of the radiotap to 8 bytes from the start of the buffer. I am not sure that is a proper solution.

Moreover, I cannot see any frames being discarded anymore.

### DIFF
--- a/include/tins/radiotap.h
+++ b/include/tins/radiotap.h
@@ -366,6 +366,7 @@ namespace Tins {
                     dbm_signal:1,
                     dbm_noise:1,
                     lock_quality:1,
+
                     tx_attenuation:1,
                     db_tx_attenuation:1,
                     dbm_tx_power:1,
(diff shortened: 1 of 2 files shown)


haralduna:
The fix may be ok wrt to radiotap header, but it breaks further processing of most frames .... I guess the alignment operation is no good.

mfontanini:
Thanks for fixing the field ordering. Yesterday I didn't have much time, so I just managed to setup the cross-compiling environment. Today I'll put my hands on this for sure.

haralduna:
It is  working for me with the last fix, on both Yun and Ubuntu.
Feel free to rework or improve the solution. There are also some comments (commented out) that should be removed.

mfontanini:
Closing since this was fixed on another PR.

PR #63, opened 2015-03-31
Fixed field name corresponding to IEEE 802.11-2012

UlfWetzker:
According to the IEEE802.11 2012 Standard (8.4.1.4 Capability Information field) the flag named reserved is now called "Radio Measurement".


mfontanini:
Thanks for the patch!

Issue #64, opened 2015-04-06
Compilation fails on VS2013


blazeable:
I'm trying to compile the latest sources using VS2013 targeting x64, but I'm getting compilation errors in include/tins/internals.h. VS2013 doesn't like this template:
`template <class T, class P, class=void>`
`struct accepts_type : std::false_type { };`

`template <class T, class P>`
`struct accepts_type<T, P, `
`    typename std::enable_if<`
`        std::is_same< decltype(  std::declval<T>()(std::declval<P>())  ), bool>::value`
`    >::type `
`> : std::true_type { };`

Which would be a bug (or lack of support) in VS2013, as it compiles fine elsewhere.

However, I'm getting this even when configuring with LIBTINS_ENABLE_CXX11=0. It seems the TINS_IS_CXX11 definition in cxxstd.h does not check the HAVE_CXX11 macro from config.h.

mfontanini:
That shouldn't happen if you set that option to 0. Remember that CMake
caches options, so if you set it to 1 and then to 0, it won't work. Have
you tried that on a clean compilstion?

As for the error. I'll see if there's any workarounds.

Thanks for the info.

mfontanini:
Sorry, you're right. The switch doesn't do anything on VS, since VS uses C++11 by default. I'll add a quick fix now, just disabling that section of code, and try to fix it tonight.

mfontanini:
Okay, I've just pushed a fix that disables that part of the code when using Visual Studio, so you should now be able to compile it. 

Sorry for the troubles. I'll have a look at this later, although I just tried a few things and apparently VS doesn't like the use of ```decltype``` and ```declval``` there. Hopefully there's some workaround for this. I'll keep you updated.

blazeable:
Thanks for the fix, I was able to build the libtins library with LIBTINS_ENABLE_CXX11=1 defined :) However I'm getting compilation errors in the example projects, seems the build system isn't adding the pcap and openssl directories to the include path for some of the generated VS projects. Looking into it.

c0mradee:
> I'm trying to compile the latest sources using VS2013 targeting x64, but I'm getting compilation errors in include/tins/internals.h. VS2013 doesn't like this template:
> `template <class T, class P, class=void>`
> `struct accepts_type : std::false_type { };`
> 
> `template <class T, class P>`
> `struct accepts_type<T, P,`
> `typename std::enable_if<`
[...]

PR #65, opened 2015-04-06
Ensure HAVE_CXX11 is checked when defining TINS_IS_CXX11

blazeable:
This is how I fixed mfontanini/libtins#64, enabling compilation on VS2013 when configuring with LIBTINS_ENABLE_CXX11=0 defined. Still seems like a good idea to check this macro here, even with mfontanini/libtins@3478c7b09a746b5e3fbcbbf2e84e606a7291e2a5.

### DIFF
--- a/include/tins/cxxstd.h
+++ b/include/tins/cxxstd.h
@@ -30,6 +30,8 @@
 #ifndef TINS_CXXSTD_H
 #define TINS_CXXSTD_H
 
+#include "config.h"
+
 #include <memory>
 
 #ifdef __GXX_EXPERIMENTAL_CXX0X__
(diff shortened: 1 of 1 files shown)


mfontanini:
Can you include config.h in macros.h? Otherwise you need to have included config.h for this macro to work correctly.

blazeable:
Yes, you're right! Didn't realize that before. Also, using 'defined(x)' isn't supposed to work elsewhere than in an 'if' preprocessor directive, so I updated the previous change (also prevents C4668 in VS2013).

blazeable:
Another commit to make sure TINS_IS_CX11 can't be redefined.

mfontanini:
Sorry, I thought the CXX11 definitions were in macros.h. So the config.h include line should be in cxxstd.h rather than in macros.h.

blazeable:
Done.

mfontanini:
Seems good, thanks!

PR #66, opened 2015-04-08
Made rsn_information() a const member function 

UlfWetzker:
Made rsn_information() a const member function in order to make Dot11ManagementFrame immutable. 

Otherwise a non const rsn_information() called like this:

    Tins::RSNInformation rsn_info = tins_mgmt_frame->rsn_information();

will result in following compiler error:

    candidate function not viable: no known conversion from 'const Tins::Dot11AssocRequest' to
    'Tins::Dot11ManagementFrame' for object argument RSNInformation rsn_information();

### DIFF
--- a/include/tins/dot11/dot11_mgmt.h
+++ b/include/tins/dot11/dot11_mgmt.h
@@ -834,7 +834,7 @@ public:
      * 
      * \return std::string containing the ssid.
      */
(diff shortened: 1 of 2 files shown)


mfontanini:
Yep, that's correct. Thanks!

PR #67, opened 2015-04-09
Added channel_map_type

UlfWetzker:
I noticed that channels_type has two different meanings.
It is used in combination with the supported_channels() method and it is part of the ibss_dfs_params struct used in combination with the ibss_dfs() method. Since channels_type has a rather general definition it will work without any problems.

    /**
     * The supported channels container type.
     */
    typedef std::vector<std::pair<uint8_t, uint8_t> > channels_type;

But without looking into the IEEE 802.11 standard I got quite confused by this dual use.

Actually the difference is pretty significant.
In section 8.4.2.20 "Supported Channels element" of IEEE 802.11-2012 channels_type would have the following content:

    channels_type.first == "First Channel Number" field -> integer
    channels_type.second == "Number of Channels" field -> integer

Within the Context of the IBSS DFS as it is described in section 8.4.2.26 "IBSS DFS element" channels_type would represent the channel map field :

    channels_type.first == "Channel Number" field -> integer
    channels_type.second == "Map" -> flags

To avoid any possibility of confusion I introduced an identical typedef called channel_map_type.
I think there are no drawbacks by doing so.

In addition a created an enumeration called MapMask to set the flags of the map field in a convenient way.

### DIFF
--- a/include/tins/dot11/dot11_mgmt.h
+++ b/include/tins/dot11/dot11_mgmt.h
@@ -71,8 +71,8 @@ public:
     /**
      * \brief Enum used in the reason code field.
      *
-     * These  can be used to get or set the second value of
-     * ibss_dfs_params().channel_map
+     * This enumeration can be used to get or set the reason code field in a
+     * Deauthentication or Disassociation
      */
     enum ReasonCodes {
         UNSPECIFIED = 1,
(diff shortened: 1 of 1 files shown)


mfontanini:
Thanks!

PR #70, opened 2015-04-20
Perserve IP protocol when using RawPDU

benmcmorran:
Currently, the IP protocol field for non-fragmented packets is forcibly overwritten to the protocol of the inner PDU even if that PDU is a RawPDU. I believe this behavior contradicts the documentation.

> If this packet has been crafted manually and the inner_pdu is, for example, a RawPDU, then setting the protocol yourself is necessary.

I discovered this when trying to serialize an IP PDU with an inner RawPDU. Despite explicitly setting the protocol of the IP PDU before serialization, it always serialized as `0xff`.

### DIFF
--- a/src/ip.cpp
+++ b/src/ip.cpp
@@ -385,7 +385,7 @@ void IP::write_serialization(uint8_t *buffer, uint32_t total_sz, const PDU* pare
                 Internals::pdu_type_to_id<IP>(inner_pdu()->pdu_type())
             );
         }
-        if(!is_fragmented() || new_flag != 0xff)
+        if(!is_fragmented() && new_flag != 0xff)
             protocol(new_flag);
     }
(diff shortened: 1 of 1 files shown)


mfontanini:
While I agree with the code in your initial commit (which I had a look at this morning but have just tried the tests and made sure they work), you have now pushed a new commit to the same branch which breaks a test and I don't think that it belongs here. That's why you should create a separate branch for pull requests, so that they don't get mixed up with newer, unrelated commits.

If you roll back that last commit and only leave bc4aea6, then I'll merge this PR. You could also create another branch and PR and close this one.

benmcmorran:
Whoops, sorry about that. It should be good now.

mfontanini:
Thank you :D

PR #78, opened 2015-05-18
Getter and Setter for handshakes(ptk)

prasad24582:
Hi Matias, 
This pull request for to add getter and setter for handshakes, its not exactly sets/gets handshakes. This pull request will allow us to set/get ptk (after processing handshakes). 

Changed spaces and way of return as per your mail. If still problem is there, please let me know, will do it. Am sorry for this because this is first time am working on git(creating pull request).

Please let me know if any change required.

Thanks and Regards,
Prasad

### DIFF
--- a/include/tins/crypto.h
+++ b/include/tins/crypto.h
@@ -157,7 +157,7 @@ namespace Crypto {
          * failed, true otherwise.
          */
         bool decrypt(PDU &pdu);
-    private:
+    protected:
         typedef std::map<address_type, std::string> passwords_type;
     
         PDU *decrypt(RawPDU &raw, const std::string &password);
--- a/include/tins/crypto.h
+++ b/include/tins/crypto.h
@@ -66,18 +66,11 @@ namespace Crypto {
             
             SessionKeys();
             SessionKeys(const RSNHandshake &hs, const pmk_type &pmk);
-            SessionKeys(const ptk_type &rptk, const bool &ccmp) {
-                std::copy( rptk.begin(), rptk.end(), ptk.begin());
-                is_ccmp = ccmp; 
-            }
+            SessionKeys(const ptk_type &rptk, const bool &ccmp); 
 
             SNAP *decrypt_unicast(const Dot11Data &dot11, RawPDU &raw) const;
-            const ptk_type &get_ptk() const {
-                return ptk;
-            }
-            bool get_ccmp() const {
-                return is_ccmp;
-            }
+            const ptk_type &get_ptk() const;
+            bool get_ccmp() const;
 
         private:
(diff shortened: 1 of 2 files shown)


mfontanini:
I think that if the goal is to be able to add a handshake from outside of the class, rather than deriving, it would be just better to allow doing that from the outside, by, for example, making ```try_add_keys``` public. 

I wouldn't complain about making this ```decrypt``` virtual and allowing to derive this class, but it seems like it would be just a hack to be able to inject something else. Rather than doing that, I would just allow doing what you'd be currently doing on your overriden implementation.

Can you give some examples of what you'd try to achieve apart from just adding session keys? For example, you shouldn't even need the handshake capturer that's inside the ```WPA2Decrypter``` for that. You could just create another ```HandshakeCapturer``` outside, feed it your packets and then extract the handshake and use it on ```WPA2Decrypter::try_add_keys``` (which would now be public).

prasad24582:
let me explain you in brief about my project, we are sniffing wireless data on a specific network and trying to decrypt packets. While decrypting, sniffer application may restart many times. So, we are looking for adding previous captured handshakes. In this way, no need to send De-authentication many times until unless if unable to decrypt packets.
 
Another point is that, the reason why we are trying to override decrypt is, in decrypt if Sniffer application not able to decrypt Dot11Data packets then need to report it application.

Coming to point, if we make try_add_keys as public its really good however we need to overload that function to pass ptk_type, addr_pair and ccmp. As per current application also, if we don't want to overload try_add_keys then problem is that need to capture entire handshakes raw packet. To save entire handshakes raw packet, I think, we need to change "RC4EAPOL::RC4EAPOL(const uint8_t *buffer, uint32_t total_sz)" to save handshakes in raw packets format. That is the reason, I preferred to save "ptk_type, addr_pair and ccmp" and apply them before decryption start on a specific device. [I am not sure, there is another way to capture handshakes in raw packets format to save them into file otherwise need to Archive RSNHandshake and save it into file, I didn't try this one. Am not how good it is saving a complex object into a file. ]. 

Please let me know if you have any idea to save successful handshakes into file without changing "RC4EAPOL::RC4EAPOL(const uint8_t *buffer, uint32_t total_sz)" .

mfontanini:
Wouldn't it be better to just add a get_pmks (which returns the pmks_map) and a get keys (which returns the keys_map). Using that, you can get the pmks and the session keys (and then you could ask the session keys for its pmk and whether it's ccmp). 

So right before restarting your sniffer, you can ask for that information and then whatever you want with it (like writing it to a file).

mfontanini:
And of course, making try_add_keys public as I said before, so you can set those keys back after you restart the capture.

prasad24582:
will pull as per your comments.

PR #80, opened 2015-05-21
getter and setters for ptk, ccmp, keys, pmks, capturer and aps

prasad24582:
Hi Matias,

In this pull request added extract_addr_pair and extract_addr_pair_dst as public because while trying to use them in derived class creating access violation, to get rid of that access violation again we need to implement them in derived. Didn't see any reason to avoid them in public along with getter and setters for ptk, ccmp, keys, pmks, capturer and aps. If you feels that's not a good idea please let me know, will remove them from pull request.

Thanks and Regards,
Prasad

### DIFF
--- a/include/tins/crypto.h
+++ b/include/tins/crypto.h
@@ -283,6 +283,26 @@ namespace Crypto {
         void add_access_point(const std::string &ssid, const address_type &addr);
         
         /**
+         * \brief extract address pair of access point and device bssid pair 
+         * by extracting data from dot11 packet.
+         * 
+         * \param dot11.
+         * 
+         * \returns pair of access point and device bssid.
+         */
+        addr_pair extract_addr_pair(const Dot11Data &dot11);
+        
+        /**
+         * \brief extract address pair of access point and device bssid pair 
+         * by extracting data from dot11 packet.
+         * 
+         * \param dot11.
+         * 
+         * \returns pair of access point and device bssid.
+         */
+        addr_pair extract_addr_pair_dst(const Dot11Data &dot11);
+        
+        /**
          * \brief get pmks.
          * 
          * return pmks.
@@ -325,8 +345,6 @@ namespace Crypto {
                 std::make_pair(addr1, addr2) :
                 std::make_pair(addr2, addr1);
         }
-        addr_pair extract_addr_pair(const Dot11Data &dot11);
-        addr_pair extract_addr_pair_dst(const Dot11Data &dot11);
         bssids_map::const_iterator find_ap(const Dot11Data &dot11);
 
         RSNHandshakeCapturer capturer;
--- a/include/tins/crypto.h
+++ b/include/tins/crypto.h
@@ -66,7 +66,7 @@ namespace Crypto {
             
             SessionKeys();
             SessionKeys(const RSNHandshake &hs, const pmk_type &pmk);
-            SessionKeys(const ptk_type &rptk, const bool &ccmp);
+            SessionKeys(const ptk_type &rptk, const bool ccmp);
             SNAP *decrypt_unicast(const Dot11Data &dot11, RawPDU &raw) const;
             const ptk_type &get_ptk() const {
                 return ptk;
--- a/include/tins/crypto.h
+++ b/include/tins/crypto.h
@@ -260,7 +260,7 @@ namespace Crypto {
          */
         bool decrypt(PDU &pdu);
         
-         /**
+        /**
          * \brief add handshakes(keys).
          * 
          * A Dot11Data PDU is looked up inside the provided PDU chain.
@@ -272,7 +272,7 @@ namespace Crypto {
          */
         void try_add_keys(const Dot11Data &dot11, const RSNHandshake &hs);
         
-         /**
+        /**
          * \brief add access point ssid and address.
          * 
          * Add ssid to access point to bssids map if ssid exits in pmks map.
@@ -288,7 +288,7 @@ namespace Crypto {
          * return pmks.
          */
         pmks_map &get_pmks() {
-          return pmks;
+            return pmks;
         }
         
         /**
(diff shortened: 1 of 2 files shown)


mfontanini:
Why are you exposing everything on the WPA2Decryptor now? I don't see a reason to expose the bssid_map since the user is responsible for filling that map. I also don't see the point in exposing the handshake capturer. Specially exposing non const references to them. And why is now ```extract_addr_pair``` public? You don't think you should have to use those functions. I thought we agreed on the fact that this class shouldn't be derived. That was the whole point of making some methods public.

Tonight I'm going to have some time so I'm going to fix this. I agree that WPA2Decryptor is a big black box and that should be modified.

prasad24582:
about bssid_map(crypto.cpp), in decrypt function, am not sure how to achieve following functionality without adding additional bssid_map in derived class.

 if(aps.count(beacon->addr3()) == 0) {
            try {
                std::string ssid = beacon->ssid();
                if(pmks.count(ssid)) {
                    add_access_point(ssid, beacon->addr3());
                }
            }
            catch(option_not_found&) { }
        }

The same thing happens with "extract_addr_pair" in the following code(crypto.cpp), so exposed extract_addr_pair as public otherwise need to re-implement (extract_addr_pair and extract_addr_pair_dst) in derived class.

 if(data && raw && data->wep()) {
            // search for the tuple (bssid, src_addr)
            keys_map::const_iterator it = keys.find(extract_addr_pair(*data));

            // search for the tuple (bssid, dst_addr) if the above didn't work
            if(it == keys.end())
                it = keys.find(extract_addr_pair_dst(*data));
            if(it != keys.end()) {
                SNAP *snap = it->second.decrypt_unicast(*data, *raw);
                if(snap) {
                    data->inner_pdu(snap);
                    data->wep(0);
                    return true;
                }
            }

>>Tonight I'm going to have some time so I'm going to fix this. I agree that WPA2Decryptor is a big black box and that should be modified. 
That would be great if you are going to change code such way like to make WPA2Decryptor derivable. 

Thank you.

mfontanini:
The bssid code is the same as calling ```add_ap_data``` using both the ssid and the bssid, so that's covered.

So you're basically reimplementing decrypt on the derived class? Because it looks like that's what you're doing. We said there was going to be no derived class, so I don't think this applies here. 

I'll try to fix this tonight.

prasad24582:
you are right about decrypt function, my ultimate goal is that in derived class should able to achieve following functionality:

bool decrypt(PDU &pdu){
if(capturer.process_packet(pdu)) {
        try_add_keys(pdu.rfind_pdu<Dot11Data>(), capturer.handshakes().front());
        //send successful keys(ptk, ccmp, addr pair) to calling application here, so that calling application can save them
        capturer.clear_handshakes();
    }
    ...... /*original code*/

}

In the existed code to check new successful device, After every packet processed need to check keys map for new successful key after calling decrypt function(bool decrypt(PDU &pdu), which I don't like because "capturer.process_packet(pdu)" doing what I need. 

Set saved keys:
WPA2DecryperDerived::SetKeys(ptk, ccmp, addr_pair) {
  keys[addr_pair] = Session(ptk, ccmp);
}

Because of above two above reasons only, am trying to write derived class. Otherwise, am pretty much happy to use the existing code.

mfontanini:
Aha, great, I see. I'll take this into account. Will update you on this tonight.

prasad24582:
Thank you Matias.

mfontanini:
Okay, I fixed this. I had to change a few things to allow this (like SessionKeys was using something on the Internals namespace, which shouldn't be exposed). You can have a look at the [WPA2Decrypter test](https://github.com/mfontanini/libtins/blob/master/tests/src/wpa2_decrypt.cpp#L141) where you can see how it's being used. 

You would basically need to check if there are new keys after every packet. It's not the greatest way to do it, but it's not bad either. 

Let me know if this works for you. Sorry I didn't merge the pull request, there were many changes that I wanted to make sure were right (like changing those types).

prasad24582:
Thank you Matias, I will check and let you know how your changes work with me. I think, its better to keep WPA2Decrypter derivable.

mfontanini:
I don't agree on the derivability (?) of WPA2Decrypter. It's a class with *one* purpose: decrypting WPA2 traffic. And that's done in just one method because most of the code is scattered around in different classes (SupplicantData, SessionKeys, etc). Letting people override that one method seems like making the base class useless.

If WPA2Decrypter really doesn't do what you need, you can always use the same code (which again, is really short, since the actual decryption code is on other classes which you can still use) on another class that best fits your needs.

prasad24582:
>>I don't agree on the derivability (?) of WPA2Decrypter. It's a class with one purpose: decrypting WPA2 traffic. And that's done in just one method because most of the code is scattered around in different classes (SupplicantData, SessionKeys, etc). Letting people override that one method seems like making the base class useless.

Agree with you. 

>>If WPA2Decrypter really doesn't do what you need, you can always use the same code (which again, is really short, since the actual decryption code is on other classes which you can still use) on another class that best fits your needs

moving  try_add_keys and add_access_point as public helps me. is it possible? 
Sorry for asking like this.

mfontanini:
You shouldn't need ```add_access_point``` to be public. Calling [this overload](https://github.com/mfontanini/libtins/blob/master/include/tins/crypto.h#L294) is equivalent, and it's already public.

On the other hand, why do you need ```try_add_keys``` to be public? [```add_decryption_keys```](https://github.com/mfontanini/libtins/blob/master/include/tins/crypto.h#L316) does exactly the same, but it's even simpler since you only need the client/ap addresses rather than a whole packet (and you were extracting this from another WPA2Decrypters keys, right? So you should have them). If the problem is that you have a RSNHandshake and that's why you want it to be public, you can still use ```add_decryption_keys```, since a ```SessionKeys``` [can be constructed from a ```RSNHandshake```](https://github.com/mfontanini/libtins/blob/master/include/tins/crypto.h#L100). You only need the PMK for this, but that can be generated by using [```SupplicantData```](https://github.com/mfontanini/libtins/blob/master/include/tins/crypto.h#L140). That class is not documented since it was supposed to be used internally, but I'll add the docs this weekend.

prasad24582:
Thank you Matias for your explanation. Will use them as you suggested. Am closing this request.

Issue #89, opened 2015-07-13
Bug in DHCPv6::add_option(const option &opt)


bkambach:
There is a bug in DHCPv6::add_option(const option &opt), such that the length of options is not accounted for.  This results in "malloc(): memory corruption" errors.

The following patch makes it better, but I'm not sure this is the best solution (namely the "+ 4"):
```
diff --git a/src/dhcpv6.cpp b/src/dhcpv6.cpp
index e5f8d47..a8079df 100644
--- a/src/dhcpv6.cpp
+++ b/src/dhcpv6.cpp
@@ -82,6 +82,7 @@ DHCPv6::DHCPv6(const uint8_t *buffer, uint32_t total_sz)
     
 void DHCPv6::add_option(const option &opt) {
     options_.push_back(opt);
+    options_size += opt.data_size() + 4;
 }
 
 const DHCPv6::option *DHCPv6::search_option(OptionTypes id) const {
```

The error is generated when calling serialize() with the following packet:
```
Tins::IPv6 input_packet = Tins::IPv6("febf::1", "febf::2");
input_packet /= Tins::UDP(546, 547);
Tins::DHCPv6 payload;
payload.msg_type(Tins::DHCPv6::MessageType::REPLY);
Tins::DHCPv6::ia_na_type ia_na(1, 3, 7); // t1=3, t2=7
payload.ia_na(ia_na);
input_packet /= payload;
```

mfontanini:
Thanks for the report and the code that reproduces it! I'll have a look at this tonight. 

The only thing I'd change there would be to instead of using ```4```, use ```sizeof(uint16_t) * 2``` (that's 2 for option type and 2 for size). If you do it that way and make a PR, I'd merge it tonight (I'll also add a test for this, to make sure it doesn't happen again).

Thanks!

bkambach:
Also, is there an easier way to add an IA Address sub-option to a DHCPv6 IA_NA option?

I'm having to do:
```
Tins::IPv6 input_packet = Tins::IPv6("febf::1", "febf::2");
input_packet /= Tins::UDP(546, 547);
Tins::DHCPv6 payload;
payload.msg_type(Tins::DHCPv6::MessageType::REPLY);
// Create IA Address option as vector<uint8_t>
std::vector<uint8_t> buffer(sizeof(uint16_t) * 2 + sizeof(uint32_t) * 2 + 16);
buffer[0] = 0;
buffer[1] = 5; // IA Address
buffer[2] = 0;
[...]

PR #90, opened 2015-07-13
Increment option size when adding a new option

bkambach:
This addresses issue #89


mfontanini:
Thanks!

Seems like tests fail since the DHCPv6 options_size is being set on the constructor. I think if you remove that line, in should work. I'll have a look at this and merge it manually otherwise, no problem.

bkambach:
Forgot to run the unit tests locally.  I updated the constructor and the unit tests pass.

mfontanini:
Thanks!

PR #91, opened 2015-07-22
Incase of tunnel interface(VPN) mac address can be 0

rklabs:
I was trying to run interfaces_info.cpp from examples and I got runtime_error exception. This was happening because ifa_addr (struct ifaddrs) is NULL and hence found_hw is set to false. This is valid case and hence I'm removing the exception.

selection_001

### DIFF
--- a/src/network_interface.cpp
+++ b/src/network_interface.cpp
@@ -224,10 +224,7 @@ NetworkInterface::Info NetworkInterface::info() const {
     InterfaceInfoCollector collector(&info, iface_id, iface_name.c_str());
     info.is_up = false;
     Utils::generic_iface_loop(collector);
-    // If we didn't event get the hw address, this went wrong
-    if(!collector.found_hw) {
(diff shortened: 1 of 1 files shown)


mfontanini:
By doing this, you'd never throw, even if you didn't find the interface. I think a better fix would be something like:

```
if(!collector.found_hw && !collector.found_ip) {
    throw std::runtime_error("Error looking up interface address");
}
```

That way, you'd succeed whether you found the IP address info or the HW address. Does that work on your case?

rklabs:
Yes, you are right it's better to keep the exception in case interface does not exist. Your fix is working for me. Thanks!

mfontanini:
Great, thanks for the feedback. Can you push the fix on this same branch so I can merge your PR? :)

rklabs:
Done!

mfontanini:
Thanks!

PR #100, opened 2015-08-31
Fix absolute paths in exported CMake config file.

zmb3:
The current CONFIGURE_FILE() command produces a libtinsConfig.cmake
file with a hardcoded absolute path to the include directory.
As a result, the generated config file is not relocatable.

This is especially difficult when cross compiling where it is often
desirable to have a directory that mirrors the root filesystem of
your target (CMAKE_FIND_ROOT_PATH).  Since the config file has
absolute paths, it can't easily be shared amongst team members or
committed to version control.

This commit uses CONFIGURE_PACKAGE_CONFIG_FILE() instead, which
avoids hardcoded paths in the generated config file.

For each PATH_VAR passed in, CMake sets a PACKAGE_<var> variable
that is relative to the installed location of the package.

See http://www.cmake.org/cmake/help/v3.0/module/CMakePackageConfigHelpers.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.1)
 PROJECT(libtins)
 
+INCLUDE(CMakePackageConfigHelpers)
+
 # Compile in release mode by default
 IF(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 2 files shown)


zmb3:
Closing this for now.  I will revisit the issue soon :)

mfontanini:
Cool, thanks!

PR #107, opened 2015-10-20
Make package relocatable (fixes #101)

### LINKED ISSUE #101: Installed package not relocatable
The installed package for libtins is not relocatable (it contains a bunch of absolute paths valid only on the machine in which is was built).

There are several sub-issues here:

1. `CONFIGURE_PACKAGE_CONFIG_FILE()` should be used instead of `CONFIGURE_FILE()`
   (see https://github.com/mfontanini/libtins/pull/100)
2. The libtinsTargets-xxx.cmake file(s) that get generated contain hard coded absolute paths to   
    dependencies like libpcap.
3. The libtins.pc pkg-config file contains absolute paths on the build machine instead of paths relative
   to the install prefix.

I'm working on a PR for 1 (not quite there yet).  This will remove hard coded paths from libtinsConfig.cmake.

For 2, I believe libtins needs to IMPORT pcap instead of doing `target_link_libraries`.  See http://www.cmake.org/cmake/help/v3.2/manual/cmake-packages.7.html#creating-packages.

> Note that it is not advisable to populate any properties which may contain paths, such as `INTERFACE_INCLUDE_DIRECTORIES` and `INTERFACE_LINK_LIBRARIES`, with paths relevant to dependencies.  That would hard-code into installed packages the include directory or library paths for dependencies **as found on the machine the package was made on**.

I can't quite figure out how to fix this.  I believe instead of `target_link_libraries(tgt PCAP)` there should be a `find_dependency(PCAP)` in libtinsConfig.cmake.

I think 3 should be resolved with the same type of solution as 1.

zmb3:
These changes remove absolute paths from the package files installed by CMake.


zmb3:
Not sure why the CI builds are failing yet.  Perhaps they have an outdated version of CMake.

tfar:
AFAIK travis doesn't provide an easy way to have a newer CMake. But you should be able to download the CMake linux package from cmake.org, install it to a local prefix and adjust the PATH environment variable.

mfontanini:
I just found [this](https://docs.travis-ci.com/user/trusty-ci-environment/). Trusty has cmake `2.8.12.2` so that should work. I'll see if this doesn't have any drawbacks (just skimmed through the doc) and that it works.

There are issues with this PR though, as in, the Windows build failing because some libpcap variable is not set properly.

zmb3:
I just got this all up and running on a Windows box with VS 2015 community and CMake 3.5.2 and everything built and linked to PCAP just fine.

```
libtins\build>cmake.exe .. -G "Visual Studio 14 Win64" -DPCAP_ROOT_DIR=C:\WpdPack -DLIBTINS_BUILD_SHARED=0 -DLIBTINS_ENABLE_WPA2=0 
```

I see the appveyor.yml should be building for VS 2013 and 2015, but it I look at the failed builds I only see them running with VS 2013.

Is anyone able to reproduce these PCAP errors on Windows?

mfontanini:
I re-triggered the [build](https://ci.appveyor.com/project/mfontanini/libtins/build/1.0.265) yesterday and apparently it's happening on both vs2013 and 2015. At the time you made the initial PR, it was only building on vs2013 on appveyor.

PR #114, opened 2015-11-24
Added pcap_setdirection to sniffer

jllorente:
I have added to the sniffer the the capability of setting the capturing direction with pcap - pcap_setdirection. Now, if you can decide whether you capture all traffic or only incoming or outgoing. This can be particularly interesting for those who only need to capture packets in a single direction, as it was in my case - incoming packets only -  and could not rely on any filter of IP or MAC addresses because of my particular use case. Please, merge if if you find it interesting.

### DIFF
--- a/include/tins/sniffer.h
+++ b/include/tins/sniffer.h
@@ -47,6 +47,8 @@ namespace Tins {
     class SnifferIterator;
     class SnifferConfiguration;
 
+    typedef pcap_direction_t tins_direction_t;
+
     /**
      * \class BaseSniffer
      * \brief Base class for sniffers.
@@ -211,6 +213,14 @@ namespace Tins {
          * \param ms The amount of milliseconds.
          */
         void set_timeout(int ms);
+
+        /**
+         * \brief Sets direction for the sniffer.
+         *
+         * This calls pcap_setdirection using the provided parameter.
+         * \param d The direction for the sniffer.
+         */
+        int set_direction(tins_direction_t d);
 
         /**
          * \brief Sets whether to extract RawPDUs or fully parsed packets.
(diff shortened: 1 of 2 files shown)


mfontanini:
That looks good (travis says that failed but it seems like osx tests were failing before that). Just 2 things:

* I wouldn't typedef pcap_direction_t. Just use that type (we're already doing this for other pcap types like pcap_t).
* Do you know for how long `pcap_setdirection` has been supported? If it's fairly new, we'd need to add something like what is currently being done for `pcap_set_immediate_mode` (see [this](https://github.com/mfontanini/libtins/blob/30445f1e975b275c2246f477ba62db986c17abec/cmake/Modules/FindPCAP.cmake#L77) cmake script).

jllorente:
Thanks! I did another commit with the proposed changes. Also, I have checked the function pcap_setdirection and it seems pretty old, already available in the LIBPCAP release 1.0.0 as of October 27, 2008.

mfontanini:
Okay, just one more thing. Can you make ```set_direction``` return a ```bool``` rather than an ```int```? So that is, it would return true if ```pcap_setdirection``` returns 0, ```false``` otherwise.

jllorente:
Done!

mfontanini:
Thank you!

sgtdede:
Hi Mathias.

I've encountered issues that I never get before since this last patch on Windows platform (I'm actually working on Windows 10 but I guess it would be the same on an older one).
The exception I caught was "device or resource busy".

It seems that pcap_setdirection is not supported on winpcap. I didn't look a lot on the subject but there might be some ways to do get the same behavior on windows by giving a specific flag to pcap_open()
(cf. http://stackoverflow.com/questions/14019420/pcap-capturing-outgoing-packets).

However, the same code I am working one is actually running well on Linux, as expected.

mfontanini:
Thanks for letting me know. I've just pushed a fix for this.

I'll have a look at that later to see how to do it on Windows. For now, we just won't call pcap_setdirection at all on that platform.

sgtdede:
My pleasure :),

Ok, thanks for the fix

PR #137, opened 2016-03-16
calc the complete addressrange of a network when call AddressRange::from_mask

stubbfel:
Hi,

i was a bit confuse about the result of `AddressRange::from_mask` , because of the comment 

> Creates an address range from a base address

i thought that method return the complete network address range of the ip and subnet mask e.g: 
- 192.168.1.128/255.255.255.0 ⇒ 192.168.1.1 ... 192.168.1.254 

but the method return only the range from the base ip to end of the network (depends of subnetmask)
- 192.168.1.128/255.255.255.0 ⇒ 192.168.1.128 ... 192.168.1.254

So i added ` Internals::first_address_from_mask `- methods and add a call in `AddressRange::from_mask`  instead of pass the base ip. So method returns the range  from the first to the last ip depends of a address in the network and the subnet mask.

best regards
stubbfel

### DIFF
--- a/include/tins/address_range.h
+++ b/include/tins/address_range.h
@@ -5,14 +5,14 @@
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
  * met:
- * 
+ *
  * * Redistributions of source code must retain the above copyright
  *   notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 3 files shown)


mfontanini:
Yeah that's true, the first address should be masked as well. I'll merge it once CI tests succeed.

PR #144, opened 2016-04-12
Calculation of VLAN payload type should not override existing when UNKNOWN

CityOfSolitude:
I stumbled across a small inconsistency between serialization of EthernetII and VLAN PDUs. Both try to set the correct payload type on serialization, but in the VLAN code it just overrides whatever the type was before, whereas in the EthernetII case, a guard prevents the override if the determined type is UNKNOWN.

I adjusted the VLAN PDU code to behave the same way, as this seems more consistent (and is actually helpful to me :-))

Sorry for the noise due to whitespace changes, it is really just one tiny change...


CityOfSolitude:
Ups, accidentally based the PR on master... Should I create a new one?

mfontanini:
Thanks! No problem, both were pointing at the same commit

PR #149, opened 2016-05-26
Fixing server payloads for user buffer management

avikama:
It seems that when auto_cleanup_payloads is set to false, the server payloads are still cleaned getting cleaned.
Fixed it here, as it seems only the client gets set to false.

### DIFF
--- a/include/tins/tcp_ip/stream_follower.h
+++ b/include/tins/tcp_ip/stream_follower.h
@@ -5,14 +5,14 @@
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
  * met:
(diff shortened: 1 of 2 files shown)


mfontanini:
Oops, copy paste mistake :P. Thanks!

The commit is also adding a getter/setter for the attach to flows thing. I haven't tested that at all, so I don't really want to allow people to use it not knowing if it actually works, that's why I didn't allow people to turn that feature on.

If you only commit the payload cleanup thing, I'll merge this, but I can't merge the other part until I actually test it.

mfontanini:
Closing this as it's no longer relevant (the main issue was already addressed)

PR #152, opened 2016-06-07
Helper for creating StreamIdentifier from const Stream&

j3parker:
Does this make sense?

### DIFF
--- a/include/tins/tcp_ip/stream_identifier.h
+++ b/include/tins/tcp_ip/stream_identifier.h
@@ -45,6 +45,8 @@ class IPv6Address;
 
 namespace TCPIP {
 
+class Stream;
+
 /**
  * \brief Uniquely identifies a stream. 
  *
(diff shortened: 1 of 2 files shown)


include/tins/tcp_ip/stream_identifier.h
```diff
@@ -30,6 +30,7 @@
 #ifndef TINS_TCP_IP_STREAM_ID_H
 #define TINS_TCP_IP_STREAM_ID_H
 
+#include "stream.h"
```
mfontanini: You don't need this, you can just do a forward declaration of class `Stream` inside namespace `Tins::TCPIP` and then include this on the .cpp

mfontanini:
Except for the forward declaration thing, this looks good. If you fix that, I'll merge right away. Thanks!

include/tins/tcp_ip/stream_identifier.h
j3parker: Good call! Fixed

PR #155, opened 2016-06-24
Added cmake compiling support for MinGW

ps790:
These modificaitons allow to build Libtins on Windows with cmake directly by running 
cmake ../ -DPCAP_ROOT_DIR="PATH_TO_WpdPack" -DLIBTINS_ENABLE_WPA2=0 -DLIBTINS_BUILD_SHARED=0 -G "MinGW Makefiles"


mfontanini:
Oh this is great! Thank you very much. This was something I wanted to do for a while but never actually sat down to try and make it work.

Just one thing, could you use spaces instead of tabs? The rest of the file uses spaces here.

Thanks!

mfontanini:
Thanks!

ps790:
You're welcome!
Thanks for your great work!

Am 25.06.2016 um 03:00 schrieb Matias Fontanini:

PR #160, opened 2016-08-31
Correct typo preventing user buffer management for server TCP streams.

banburybill:
Since pull request #149 has stalled, and I've just stubbed my toe on exactly the same bug, I hope this pull request is more suitable for merging.


mfontanini:
Oooops, sorry! I totally forgot about that. Thanks for the patch!

banburybill:
And thank you too for libtins. Excellent library. We're using it here on a DNS related project that will be open sourced at some point.

PR #163, opened 2016-09-12
Feature - Skipping ahead in TCP flows.

CityOfSolitude:
I was recently using the (new?) stream follower code for TCP streams and had issues with following TCP streams of captures with packet loss. The application layer (HTTP) allowed me to "recover" from gaps in a flow, by realizing that HTTP requests and responses in a stream cannot be reordered, since a subsequent request/response will only be sent after the previous one has been completely received by the server/client, who answers with a response/request.

If packets are lost (not in the original TCP stream, but just while capturing that stream), flows will get stuck and start buffering indefinitely. By registering an out-of-order callback, the application layer can notice that gap and if it detects the start of a new response/request in an out-of-order chunk, can skip ahead in the flow, of course (maybe) discarding some partially received data in between.

I added skipTo functionality to the Flow and the data tracker to allow this. It still caused problems at the beginning of the stream, where it was not guaranteed that the data tracker process_data method was actually called after the out of order callback (the seq_compare with the initial 0 value prevented entering the execution branch and the latter out of order callback call site was used instead). I re-ordered the code a bit to allow this scenario as well.

Finally, there was a comparison of sequence numbers in Flow, which did not use seq_compare, so I fixed that as well.

(If I have the time, I will also investigate the two payload.erase calls in the data tracker, which both use a simple difference of TCP sequence numbers to determine the length of data to cut. I think this does not account for wrapping of sequence numbers: the arithmetic should follow the same RFC 1982 logic as used and implemented in seq_compare. It might be easiest to adjust seq_compare to return the right difference right away and use that for the erase calls.)

### DIFF
--- a/include/tins/tcp_ip/data_tracker.h
+++ b/include/tins/tcp_ip/data_tracker.h
@@ -108,7 +108,7 @@ public:
      *
      * \param seq The seqeunce number to skip to.
      */
-    void skipTo(uint32_t seq);
+    void advance_sequence(uint32_t seq);
 
     /**
      * Retrieves the current sequence number
--- a/include/tins/tcp_ip/data_tracker.h
+++ b/include/tins/tcp_ip/data_tracker.h
@@ -87,7 +87,30 @@ public:
      */
     bool process_payload(uint32_t seq, payload_type payload);
 
-    /** 
+    /**
+     * \brief Skip forward to a sequence number
+     *
+     * This allows to recover from packetloss, if we just do not see all packets of
+     * an original stream. This recovery can only sensibly triggered from the application
+     * layer.
+     *
+     * The method does nothing, if the sequence number is smaller or equal to the
+     * current number.
+     *
+     * This method is particularly useful to call from an out of order callback, if
+     * the application wants to skip forward to this out of order block. The application
+     * will then get the normal data callback!
+     *
+     * The method cleans the buffer from all no longer needed fragments.
+     *
+     * IMPORTANT: If you call this method with a sequence number that is not exactly a
+     * TCP fragment boundary, the flow will never recover from this.
+     *
+     * \param seq The seqeunce number to skip to.
+     */
+    void skipTo(uint32_t seq);
+
+    /**
      * Retrieves the current sequence number
      */
     uint32_t sequence_number() const;
--- a/src/tcp_ip/flow.cpp
+++ b/src/tcp_ip/flow.cpp
@@ -127,8 +127,8 @@ void Flow::process_packet(PDU& pdu) {
     }
 }
 
-void Flow::skipTo(uint32_t seq) {
-    data_tracker_.skipTo(seq);
+void Flow::advance_sequence(uint32_t seq) {
+    data_tracker_.advance_sequence(seq);
 }
 
 void Flow::update_state(const TCP& tcp) {
--- a/src/tcp_ip/flow.cpp
+++ b/src/tcp_ip/flow.cpp
@@ -110,22 +110,25 @@ void Flow::process_packet(PDU& pdu) {
     }
     const uint32_t chunk_end = tcp->seq() + raw->payload_size();
     const uint32_t current_seq = data_tracker_.sequence_number();
-    // If the end of the chunk ends after our current sequence number, process it.
-    if (seq_compare(chunk_end, current_seq) >= 0) {
-        uint32_t seq = tcp->seq();
-        // If we're going to buffer this and we have a buffering callback, execute it
-        if (seq > current_seq && on_out_of_order_callback_) {
-            on_out_of_order_callback_(*this, seq, raw->payload());
-        }
-        if (data_tracker_.process_payload(seq, move(raw->payload()))) {
-            if (on_data_callback_) {
-                on_data_callback_(*this);
-            }
+    // If the end of the chunk ends before the current sequence number or
+    // if we're going to buffer this and we have a buffering callback, execute it
+    if (seq_compare(chunk_end, current_seq) < 0 ||
+            seq_compare(tcp->seq(), current_seq) > 0){
+        if (on_out_of_order_callback_) {
+            on_out_of_order_callback_(*this, tcp->seq(), raw->payload());
         }
     }
-    else if (on_out_of_order_callback_) {
-        on_out_of_order_callback_(*this, tcp->seq(), raw->payload());
+
+    // can process either way, since it will abort immediately if not needed
+    if (data_tracker_.process_payload(tcp->seq(), move(raw->payload()))) {
+        if (on_data_callback_) {
+            on_data_callback_(*this);
+        }
     }
+}
+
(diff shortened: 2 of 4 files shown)


include/tins/tcp_ip/data_tracker.h
```diff
@@ -87,7 +87,30 @@ class TINS_API DataTracker {
...
+     * will then get the normal data callback!
+     *
+     * The method cleans the buffer from all no longer needed fragments.
+     *
+     * IMPORTANT: If you call this method with a sequence number that is not exactly a
+     * TCP fragment boundary, the flow will never recover from this.
+     *
+     * \param seq The seqeunce number to skip to.
+     */
+    void skipTo(uint32_t seq);
```
mfontanini: I would call this something like `advance_sequence`. `skipTo` seems kind of vague, as it doesn't really tell you what it's skipping.

src/tcp_ip/flow.cpp
```diff
@@ -110,24 +110,27 @@ void Flow::process_packet(PDU& pdu) {
...
-        if (seq > current_seq && on_out_of_order_callback_) {
-            on_out_of_order_callback_(*this, seq, raw->payload());
-        }
-        if (data_tracker_.process_payload(seq, move(raw->payload()))) {
-            if (on_data_callback_) {
-                on_data_callback_(*this);
-            }
+    // If the end of the chunk ends before the current sequence number or
+    // if we're going to buffer this and we have a buffering callback, execute it
+    if (seq_compare(chunk_end, current_seq) < 0 ||
```
mfontanini: If this is to handle a special case at the beginning of the stream then I think it should be more specific. This condition as it is will trigger the out of order callback even for already seen data chunks, which technically are duplicates and not OOO.

mfontanini:
Note that before this gets merged into `master` it needs to have some unit tests to make sure it works properly. If you have troubles with that, I can add them, no problem.

mfontanini:
Thanks for the PR btw!

include/tins/tcp_ip/data_tracker.h
CityOfSolitude: Sounds good :)

src/tcp_ip/flow.cpp
CityOfSolitude: By reordering I did not (hopefully) change the semantics of this part of the code. Note that the OOO callback was also called if "seq_compare(chunk_end, current_seq) < 0". I would not actually mind if we would offer a different callback for duplicates, or at least provide this information in a boolean parameter to the callback (since as it is, my callback has to make this distinction by hand). Dropping this callback might not be a good idea, since we are just adding the ability to skip ahead, so this data might no longer be "duplicate" :)

I reordered the code to combine the two cases where the OOO callback was called and make sure that after the callback the normal "data_tracker_.process_payload" is called, if applicable. The old code could call an OOO callback (from the else part) and NOT call "data_tracker_.process_payload" afterwards (which was irrelevant without the ability to skip ahead in the flow).

CityOfSolitude:
No problem, hope these changes improve the project and/or are of use to someone.

Regarding the tests; Are there any for the new stream_follower? :)
I am not sure if I have the time at the moment to write any. I would like to write a test for a TCP stream with wrapping sequence numbers, since I believe there are still errors in the existing code in this case... but I somehow can't find and example capture and did not have the time to construct one by hand :-)

mfontanini:
Yep, there are a bunch of them [here](https://github.com/mfontanini/libtins/blob/master/tests/src/tcp_ip.cpp). This type of thing is tricky enough that I wouldn't add it without adding enough tests :P.

BTW regarding what you first mentioned about `seq_compare`, the unsigned int arithmetic does what you expect here. e.g. if you do `A - B` and A is lower than B, then this will actually do what you want. You can check out some of the tests in the link above, those handle wrapping around in many different scenarios.

CityOfSolitude:
Ah, thanks for the pointer!
You are right of course with the arithmetic, somehow I already understood that for _adding_ the payload size to the current sequence number, but thought for subtracting that would somehow not work :-)

This reminds me of one observation I made when reading the Flow code; The calculation of the payload size assumes that the payload is a RawPDU behind the TCP PDU, which is true for vanilla libTins, but might not be true if one adds more protocols. Even worse, if you were to determine the payload size via tcp->inner_pdu()->size(), this would assume that all PDUs return the size of the data they were parsed from. This might be currently true, were libTins supports mostly (or even only?) binary protocols, but is no longer obvious if you include text-based protocols like HTTP. If you do not plan on preserving whitespace for such protocols (or if you do any form of normalization), the original size will be lost.

This brings me to yet another observation; The header_size method is currently const, which seems to make sense, but not necessarily for text-based protocols. If you want to serialize an HTTP PDU, you might want to add header fields like Content-Length or Content-Type automatically (or just adjust their values), which will change the header_size! But you cannot do it as a side-effect of the header_size call, you have to wait for write_serialization, which can do the changes. Unfortunately, that call is too late, since the memory for the PDU was already allocated! So with the current design, you have to calculate the final size of the PDU within header_size, without being able to change it and then do the changes within write_serialization. Maybe the prepare_for_serialize could be generalized to allow this, currently it is called after the header_size has been determined as well.

Maybe this is not the right place to discuss all this, but it just came to mind when answering :)

mfontanini:
I've opposed to adding application layer protocols to libtins in the past because I don't think they fit well. Once something goes over a TCP stream, the notion of PDU is kind of lost, especially for protocols like HTTP.

CityOfSolitude:
You are right, it does not make much sense to parse one particular network packet as a linked list of PDUs and go beyond TCP, since the payload of TCP is fragmented over several packets anyway. So the restrictions I listed do not really worry me as much in this particular scenario as well. On the other hand, SIP over UDP is a quite different story. It fits quite well and SIP might contain SDP, etc. pp.

Using stream followers one can parse SIP of over TCP quite nicely as well, since you can pretty much just wait for the stream to end and use the re-assembled payload to parse the SIP/SDP chain.

For HTTP, as you said, this is not as easy, since you can have multiple requests/responses in one flow, will have to deal with chunked transfer and so on, yet you can do it and I am currently using PDUs to represent requests/responses. I am not sure if this makes sense for the "normal" user of libTins, but for me its a case of "why not" and in my particular application I might benefit from it in the future.

I think with a moderate amount of additional work I could write an HTTP follower, that emits HTTP response and HTTP request callbacks with HTTP/CSS, HTTP/JSON, HTTP/whatever PDU chains. Again, I am not sure the "normal" user has any additional benefit from having the CSS payload of an HTTP response as a separate PDU, but in my case I might benefit from it.

I realize that the simplicity of the libTins architecture, the ease of use and the speed will all be affected by adding application layer stuff, so I did not suggest extending the vanilla libTins with such things :-)

Thanks for a great library and the work you put into it!

mfontanini:
Btw, I haven't forgotten about this PR. I just haven't had much time recently but now I'm having some extra time so I'll make sure to look into this and merge it this week.

PR #164, opened 2016-09-13
Feature - Adding user data to TCP Streams

CityOfSolitude:
An application using the (new) TCPIP stream_follower will most likely want to track some additional information per stream. This could be done in external data structures, using some kind of stream identifier. On the other hand, there is already a nice life-cycle for streams as created by libTins.

With this addition, a user can hook into new_stream_callback, to attach and initialize his own data to the stream. On subsequent callbacks, he can retrieve this data and change it via the reference. Finally, using stream_closed_callback and stream_termination_callback, he can access it one last time.

As a simple example consider following an HTTP stream, where you might want to save the last request in the stream, so you know what the current response is answering to.

### DIFF
--- a/include/tins/tcp_ip/stream.h
+++ b/include/tins/tcp_ip/stream.h
@@ -40,6 +40,7 @@
 #include <functional>
 #include <chrono>
 #include <stdint.h>
+#include <boost/any.hpp>
 #include "../macros.h"
 #include "../hw_address.h"
 #include "flow.h"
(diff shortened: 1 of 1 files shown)


mfontanini:
This is a good idea! I just have to figure out how to use boost inside appveyor, as the PR build failed.

CityOfSolitude:
Glad you like it; served me well already in one of my applications :)

mfontanini:
Can you rebase this against `develop`? I think the appveyor build should be working now using the proper boost paths.

CityOfSolitude:
Looks good :)

mfontanini:
I'll try to merge this at night as this actually needs to have some build system changes because the only code that depends on boost is the ACK tracker. The TCP streams code wasn't depending on it. I'm fine with changing this, but it needs a bit of tweaking. I'll merge this and change it myself though, don't worry. Thanks!

PR #168, opened 2016-09-22
Fix: Parsing of incomplete packets with Ethernet padding.

CityOfSolitude:
The current parsing code for EthernetII parses the header, then the next PDU indicated by the payload type, **if** there is anything left in the stream. As a consequence, the parser fails with malformed_packet, if the remainder of the packet was padding introduced as trailer on a prior serialize call.

I added a check that sees if the total size is 60 (i.e. padded ethernet) and if the remainder of the stream is all zeros, in which case I just return from the parse. This allows to parse Ethernet-only packets that were previously unparsed by libTins, even though the payload_type is still set to something like IP.

### DIFF
--- a/src/ethernetII.cpp
+++ b/src/ethernetII.cpp
@@ -81,6 +81,14 @@ EthernetII::EthernetII(const address_type& dst_hw_addr,
 EthernetII::EthernetII(const uint8_t* buffer, uint32_t total_sz) {
     InputMemoryStream stream(buffer, total_sz);
     stream.read(header_);
+
+    // Check for zero-padding for padded packets
+    if (total_sz == 60 && std::all_of(stream.pointer(),
+                                      stream.pointer() + stream.size(),
+                                      [](uint8_t x) { return x == 0; })) {
+        return;
+    }
+
     // If there's any size left
     if (stream) {
         inner_pdu(
--- a/src/ethernetII.cpp
+++ b/src/ethernetII.cpp
@@ -82,15 +82,15 @@ EthernetII::EthernetII(const uint8_t* buffer, uint32_t total_sz) {
     InputMemoryStream stream(buffer, total_sz);
     stream.read(header_);
 
-    // Check for zero-padding for padded packets
-    if (total_sz == 60 && std::all_of(stream.pointer(),
-                                      stream.pointer() + stream.size(),
-                                      [](uint8_t x) { return x == 0; })) {
-        return;
-    }
-
     // If there's any size left
     if (stream) {
+        // Check for zero-padding for padded packets
+        if (total_sz == 60 && std::all_of(stream.pointer(),
+                                          stream.pointer() + stream.size(),
+                                          [](uint8_t x) { return x == 0; })) {
+            return;
+        }
+
         inner_pdu(
             Internals::pdu_from_flag(
                 (Constants::Ethernet::e)payload_type(), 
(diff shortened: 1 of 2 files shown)


CityOfSolitude:
I am not sure that libTins wants this fixed, just to make that 100% clear, since this is obviously a corner-case use of the library.

The same effect also happens if the packet is cut after IP, since the IP parser will then try to parse the padding as its payload, etc. pp. I will look into this more and might offer a fix for IP as well.

mfontanini:
Can you provide a code/pcap example where this happens? I'm doing this and I can't seem to reproduce the issue (maybe I didn't understand the issue properly):

```c++
#include <iostream>
#include <tins/tins.h>

using namespace Tins;

int main() {
    EthernetII eth1 = EthernetII() / IP();
    auto buffer = eth1.serialize();\
    
    std::cout << eth1.size() << std::endl;
    std::cout << buffer.size() << std::endl;
    
    EthernetII eth2(buffer.data(), buffer.size());
    std::cout << eth2.size() << std::endl;
}
```

CityOfSolitude:
Hi, there are two reasons the example does not reveal the problem:
1) You need to remove an inner PDU to make the packet "incomplete" (or cut it off). Otherwise the parser will parse the while PDU chain and the simply ignore the remaining padding.
2) The way you construct the PDU chain does NOT set the payload type, this is done on serialize. So if you cut off the IP packet and then serialize, the payload type will still be 0 and no problem appears.

So the adjusted code does a serialize to set the payload type and then cuts off right after the EthernetII PDU:

```c++
#include <iostream>
#include <tins/tins.h>

using namespace Tins;

int main() {
    EthernetII eth1 = EthernetII() / IP();
    eth1.serialize(); // sets payload type
    eth1.inner_pdu(nullptr); // make it "incomplete"
    auto buffer = eth1.serialize();

    std::cout << eth1.size() << std::endl;
    std::cout << buffer.size() << std::endl;

    EthernetII eth2(buffer.data(), buffer.size());
    std::cout << eth2.size() << std::endl;
}
```

Alternative ways to do this would be to just set the payload type to 0x800 manually, instead of serializing or to load virtually any packet from a pcap and cut it off after EthernetII or IP, since the payload types will then be set by the parser already.

mfontanini:
Okay, I see the issue now. However, I don't think the bug is in the parsing code but in the serialization one. If there's no `inner_pdu`, then the next protocol flag shouldn't be set to anything in particular. e.g. your code should be the equivalent of:

```c++
EthernetII eth;
eth.serialize(); // <- output should be the same
```

The fact that the PDU had an IP layer on top of it shouldn't matter. I just pushed a [commit](https://github.com/mfontanini/libtins/commit/552006c876829089f495122d70ff92fc29992c78) that fixes this on the `develop` branch. Can you confirm the behavior is what you expect?

Thanks!

CityOfSolitude:
Well, libTins is currently "fixing" a lot of things on serialization, like checksums, payload types, etc. so it would only be logical to solve the problem by resetting the payload type once the inner PDU is gone as well. In my application, however, it would sometimes be nice to be able to parse a packet, modify it in some way and unparse it again, with more control over what is recomputed. (Especially in cases where the packet was cut off by libpcap (snaplen < wirelen), you cannot parse the packet with libtins and unparse it again, without modifying such fields. But this is obviously another story :)

Keep in mind that the same problem appears with IP and its protocol field. If the inner PDU is removed, the EthernetII padding will be (tried to be) parsed with this protocol as well. This might affect even more PDU types that fit into the 60 byte range.

mfontanini:
Yeah, that is true. I'll change the rest of the protocols later today. Thanks!

mfontanini:
Closing this as it's no longer valid.

PR #170, opened 2016-09-28
Changes to stream/flow initialization (stream attachment and sequence numbers)

CityOfSolitude:
I would like to offer/discuss these changes to TCPIP::Stream/StreamFollower/Flow:

(a) Flow sequence numbers during/after SYN/ACK.
Currently, the sequence number from the observed SYN packet is taken to initialize the data_tracker of the respective flow. The sequence number is then incremented by 1 once the flow itself sends an ACK (since the flow will not see the ACK to its own SYN). I think this is odd and means after the initial exchange of SYN, SYN/ACK, ACK, the client flow sequence number is correct, but the server flow sequence number is still off by one.

My proposed change is to handle flow sequence numbers during SYN/ACK the same way they are handled during payload processing: They are incremented directly on observing packets that contained the data. For SYN packets that means the correct sequence number of the flow is the sequence number of the packet +1 (i.e. interpret sending SYN as sending one byte). The whole change is done by changing two lines in the current code.

So is the current implementation wrong in some way? I think the current implementation will work, since even if the server is the first to send data (based on the off-by-one sequence number), the server will still ACK (since ACK is always on, with the only exception being the initial client SYN). The server flow sequence number is corrected by this automatic ACK in update_state. 

I would still recommend the change to correct the observed sequence numbers by users of the stream_follower and its callbacks. (Edit: I just realized that libTins currently has no callbacks where users could observe this; I added per-packet callbacks to streams in my application... Nevertheless, I currently think the change would conceptually make sense and streamline the implementation.)

(b) I am interested in attaching to already running streams, since the capture files I process simply do not always contain the original SYN/ACK phase. I have seen there is such a flag in the code already and exposed it via the stream_follower constructor to use it. Is this feature work in progress?

To make it work (at least so far, can't vouch for the feature completely working with this change), I had to remove the "client_flow().process_packet()" in Stream constructor. This call seems very odd to me anyway, seems to be a workaround with the intention given in the comment: "Update client flow state". This might make sense for streams created by SYN/ACK, but not if attaching to streams later on.

So I removed this call and instead always process packets. The existing code always processed packets of existing streams, so I have no idea why we would not process the initial SYN packet as well. The only other case where processing was disabled was when no stream was found an none was created (which makes sense). So I removed the bool flag and just return in this case, with the only downside that I had to copy the cleanup code for other streams...

With this change attaching seems to work (so far). I am using it together with "advance_sequence" of #163 to "recover" the flows whenever I detect the beginning of a frame of the application (e.g. HTTP).

I hope this all makes sense and I would like to know if you expect these changes to break anything. The tests are running fine, but note that I adjusted a handshake test to expect sequence number 61 instead of 60 :-)

### DIFF
--- a/src/tcp_ip/flow.cpp
+++ b/src/tcp_ip/flow.cpp
@@ -140,7 +140,6 @@ void Flow::update_state(const TCP& tcp) {
             ack_tracker_ = AckTracker(tcp.ack_seq());
         #endif // TINS_HAVE_ACK_TRACKER
         state_ = ESTABLISHED;
-        data_tracker_.sequence_number(data_tracker_.sequence_number() + 1);
     }
     else if (state_ == UNKNOWN && (tcp.flags() & TCP::SYN) != 0) {
         // This is the server's state, sending it's first SYN|ACK
@@ -148,7 +147,7 @@ void Flow::update_state(const TCP& tcp) {
             ack_tracker_ = AckTracker(tcp.ack_seq());
         #endif // TINS_HAVE_ACK_TRACKER
         state_ = SYN_SENT;
-        data_tracker_.sequence_number(tcp.seq());
+        data_tracker_.sequence_number(tcp.seq() + 1);
         const TCP::option* mss_option = tcp.search_option(TCP::MSS);
         if (mss_option) {
             mss_ = mss_option->to<uint16_t>();
--- a/src/tcp_ip/stream.cpp
+++ b/src/tcp_ip/stream.cpp
@@ -62,8 +62,6 @@ Stream::Stream(PDU& packet, const timestamp_type& ts)
 : client_flow_(extract_client_flow(packet)),
   server_flow_(extract_server_flow(packet)), create_time_(ts), 
   last_seen_(ts), auto_cleanup_client_(true), auto_cleanup_server_(true) {
-    // Update client flow state
-    client_flow().process_packet(packet);
     const EthernetII* eth = packet.find_pdu<EthernetII>();
     if (eth) {
         client_hw_addr_ = eth->src_addr();
--- a/src/tcp_ip/stream_follower.cpp
+++ b/src/tcp_ip/stream_follower.cpp
@@ -90,7 +90,6 @@ void StreamFollower::process_packet(PDU& packet, const timestamp_type& ts) {
     }
     stream_id identifier = stream_id::make_identifier(packet);
     streams_type::iterator iter = streams_.find(identifier);
-    bool process = true;
     if (iter == streams_.end()) {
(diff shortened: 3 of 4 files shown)


mfontanini:
I never really liked the way the server flow handled sequence numbers in a different way from the client one at the beginning of the stream. This looks good!

Regarding attaching to flows, that's something I thought about adding initially (hence the flag) but never got to it. The only thing about that is writing proper tests to make sure it works as intended, as the state of the stream could be kind of unexpected at some times. e.g. you attach on an out of order packet (not really that weird and maybe it just works but we need to make sure it's handled properly).

CityOfSolitude:
If you attach to a re-transmitted packet, the flow will get stuck :)
Without any form of recovery by detecting a frame boundary in an out-of-order callback, the buffer will just fill up. You get into the same issues when you have packet loss while observing traffic, of course.

(Note that recovery by out-of-order callbacks is only a good idea if the client and server flows follow a strict request/response dialog cycle, since otherwise the requests on the client flow and the responses on the server flow could overtake each other.)

I have no idea what libTins could offer as a generic solution to all this (besided allowing the user to fix things :-)). Timeouts would be awkward, I guess.

mfontanini:
Hey! I finally sat down and did some work on this. So now you can enable the attach to flows flag on StreamFollower, I added tests and all :).

I also added an automatic way of recovering a stream (Stream::enable_recovery_mode) which can be used if there's high packet loss, but mostly for when you attach to a partial stream. This basically takes a recovery window size and overrides the client/server out of order callbacks with a function that will call Flow::advance_sequence if a packet with a sequence number if the sequence number is inside the range `(sequence number when recovery mode was enabled, that sequence number + recovery windo size]`. This is a simple mechanism that should work, considering a partial stream will be in a kind of delicate state on the first few packets. 

So basically now the new stream callback could do something like:

```c++
void on_new_stream(Stream& stream) {
    // common stuf...

    // Is this a stream we attached to mid way?
    if (stream.is_partial_stream()) {
        // Let's skip out of order packets for the first 10kb of this stream.
        stream.enable_recovery_mode(10 * 1024);
    }
}
```

Some feedback on this would be nice, especially considering you seem to have been attaching to streams already.

CityOfSolitude:
Hi,

sorry for the very late answer!

I think offering a recovery based on a configurable window size makes a lot of sense. For attaching to a stream, where retransmits are a problem, your solution should work well.

To deal with packet loss mid-stream, however, it might make more sense to look for packets outside of the window, i.e., to define a maximum time to wait to fill in blanks. Once a packet arrives with a larger number than currently reassembled + window size, we skip over the gap starting at the currently reassembled sequence number (so the window can slide forward).

PR #177, opened 2016-11-30
Update dns.cpp

shshzi:


src/dns.cpp
```diff
@@ -52,7 +52,7 @@ using Tins::Memory::OutputMemoryStream;
 namespace Tins {
 
 PDU::metadata DNS::extract_metadata(const uint8_t *buffer, uint32_t total_sz) {
-    if (TINS_UNLIKELY(sizeof(dns_header))) {
+    if (TINS_UNLIKELY(total_tz < sizeof(dns_header))) {
```
mfontanini: Good catch, but I think you meant `total_sz` here.

mfontanini:
I fixed this myself on [this commit](https://github.com/mfontanini/libtins/commit/94e5ac21095b4b2d373382313b673e128843dba9)

PR #179, opened 2017-01-05
Use actual payload length to construct inner PDU. Fixes #178.

### LINKED ISSUE #178: Malformed packet on ICMPv6 neighbor/router solicitation with EthernetII FCS
The following packet (Wireshark hexdump format) is an EthernetII frame with FCS at end, containing an IPv6 packet containing an ICMPv6 neighbor solicitation:

```
0000  33 33 ff 01 31 3e 64 3f  5f 01 31 3e 86 dd 60 00   33..1>d? _.1>..`.
0010  00 00 00 18 3a ff 00 00  00 00 00 00 00 00 00 00   ....:... ........
0020  00 00 00 00 00 00 ff 02  00 00 00 00 00 00 00 00   ........ ........
0030  00 01 ff 01 31 3e 87 00  55 69 00 00 00 00 fe 80   ....1>.. Ui......
0040  00 00 00 00 00 00 66 3f  5f ff fe 01 31 3e 23 0c   ......f? _...1>#.
0050  57 b7                                              W.               
```

When `IPv6::IPv6` constructs the inner PDU, it passes the remaining stream size (28, including the FCS) instead of the IPv6 payload length (24); this means that `ICMPv6::parse_options` tries to parse the FCS as an ICMPv6 option and throws `malformed_packet`.

This potentially affects any protocol depending on the IPv6 payload length i.e. not having its own length field, esp. TCP.

One possible fix would be to pass the IPv6 payload length (minus any extension headers) to inner PDU constructors in place of the remaining stream size.

ecatmur:


src/ipv6.cpp
```diff
@@ -93,24 +94,48 @@ IPv6::IPv6(const uint8_t* buffer, uint32_t total_sz)
...
             const uint32_t ext_size = (static_cast<uint32_t>(stream.read<uint8_t>()) + 1) * 8;
             const uint32_t payload_size = ext_size - sizeof(uint8_t) * 2;
-            if (!stream.can_read(ext_size)) { 
+            if (!stream.can_read(payload_size)) {
                 throw malformed_packet();
             }
             // minus one, from the size field
             add_ext_header(ext_header(ext_type, payload_size, stream.pointer()));
+            if (actual_payload_length == 0u && current_header == HOP_BY_HOP)
+            {
```
mfontanini: Minor syntax issue: can you move this brace up to the end of the previous line so it looks like this?

```c++
if (foo) {

}
``

File: src/ipv6.cpp [hunk 2]
```diff
@@ -93,24 +94,48 @@ IPv6::IPv6(const uint8_t* buffer, uint32_t total_sz)
...
+                // could be a jumbogram, look for Jumbo Payload Option
+                InputMemoryStream options(stream.pointer(), payload_size);
+                while (options) {
+                    const uint8_t opt_type = options.read<uint8_t>();
+                    if (opt_type == PAD_1)
+                        continue;
```
mfontanini: Can you please add braces around this statement?

src/ipv6.cpp
```diff
@@ -93,24 +94,48 @@ IPv6::IPv6(const uint8_t* buffer, uint32_t total_sz)
...
+                    const uint8_t opt_size = options.read<uint8_t>();
+                    if (opt_type == JUMBO_PAYLOAD)
+                    {
```
mfontanini: Same for this brace

src/ipv6.cpp
```diff
@@ -93,24 +94,48 @@ IPv6::IPv6(const uint8_t* buffer, uint32_t total_sz)
...
+                        if (opt_size != 4) {
+                            throw malformed_packet();
+                        }
+                        actual_payload_length = Endian::be_to_host<uint32_t>(stream.read<uint32_t>());
```
mfontanini: You can actually just do:

```c++
actual_payload_length = stream.read_be<uint32_t>();
```

tests/src/ipv6_test.cpp
```diff
@@ -175,6 +185,16 @@ TEST_F(IPv6Test, Broken1) {
     );
 }
 
+TEST_F(IPv6Test, FCSSuffix) {
```
mfontanini: Awesome that you added tests for this, thanks!

mfontanini:
Thank you!

PR #184, opened 2017-01-25
Fix -Wextra compiler warnings.

### LINKED ISSUE #183: Lots of warnings when using libtins in a project with -Wextra
libtins has several unused parameters throughout its codebase, as well some ignored qualifiers, that show up when building against it using -Wextra. It'd be great if these warnings were at least fixed in its headers, if not throughout its codebase.

kyrofa:
This PR fixes #183 by introducing an empty function template called `unused` into the `Internals` namespace and calling it on unused variables throughout the codebase. It also fixes one or two ignored qualifiers. Finally, it adds -Wextra to the build flags.

I'm happy to pair this down to only the changes you want, but as mentioned in the bug it'd be great if these warnings were at least fixed in the headers.


mfontanini:
As far as I understand the issue, would like to fix it and appreciate your PR, I don't think this is the right approach. I'd much rather either remove the parameter names on functions that don't use them or instead comment the names out like:

```c++
bool matches_response(const uint8_t* /*ptr*/, uint32_t /*total_sz*/) const { }
```

In the few cases where a parameter is only used depending on the platform, I agree with doing something like `Internals::unused` but on any other cases it just seems to add unnecessary bloat to the function.

kyrofa:
I do agree, though you use doxygen comments through here; commenting out the variable names will likely make the generated documentation less useful. Are you still okay with that?

mfontanini:
That is a good point actually. Although there's only a few of these issues in header files (all of them on empty functions, except `PDU::find_pdu` which should actually be forwarding `type` to its call, so that's a bug), so I'd rather move them into their respective cpp files, thus removing the impact on doxygen docs. All of those are virtual functions anyway, so leaving them on the header files doesn't improve anything as they won't be inlined (and they were initially defined there probably due to laziness).

I can make these changes later today if you don't feel like it.

kyrofa:
> except PDU::find_pdu which should actually be forwarding type to its call, so that's a bug)

Heh, I wondered about that. Want me to log a new bug, or just fix it here?

> so I'd rather move them into their respective cpp files

You got it, I'm on it.

mfontanini:
I would just fix it there. It's actually a bug but I don't imagine anyone passes the pdu_flag manually as it's kind of pointless.

Thanks!

kyrofa:
Done. As you noted, I left the platform-specific uses using `Internals::unused()`. Let me know what you think.

mfontanini:
Nice, thanks! Actually, there's still a few warnings, at least that I can see [on travis](https://travis-ci.org/mfontanini/libtins/jobs/195046575) by expanding `make tests`.

kyrofa:
Ah indeed, I'll take a look.

kyrofa:
Should be better now... let's see what CI thinks.

kyrofa:
Looks like appveyor needs a re-run:

> appveyor DownloadFile http://www.winpcap.org/install/bin/WpdPack_4_1_2.zip
> Error downloading remote file: One or more errors occurred.
> Inner Exception: Remote server returned 523: Origin Unreachable

include/tins/ip_reassembler.h
```diff
@@ -190,7 +190,6 @@ class TINS_API IPv4Reassembler {
     address_pair make_address_pair(IPv4Address addr1, IPv4Address addr2) const;
     
     streams_type streams_;
-    OverlappingTechnique technique_;
```
mfontanini: This one can't be removed actually. I know it's not being used but it was added so that eventually (potentially) the code can handle IP reassembly using different techniques. So this warning we'll have to cope with (plus, this should only be a warning when building libtins itself).

src/ip_reassembler.cpp
```diff
@@ -94,13 +94,11 @@ uint16_t IPv4Stream::extract_offset(const IP* ip) {
 
 } // Internals
 
-IPv4Reassembler::IPv4Reassembler()
-: technique_(NONE) {
```
mfontanini: These ones needs to be added back as stated on the other comment.

src/packet_writer.cpp
```diff
@@ -71,11 +72,11 @@ void PacketWriter::write(Packet& packet) {
 
 void PacketWriter::write(PDU& pdu, const struct timeval& tv) {
     PDU::serialization_type buffer = pdu.serialize();
-    struct pcap_pkthdr header = {
```
mfontanini: Was it actually complaining about something here?

mfontanini:
Thanks for the fixes! Appveyor only failed on one node because it was broken, but it's not going to fail for sure. Just added a few comments and then it should be good to be merged.

src/packet_writer.cpp
kyrofa: Yeah the osx Travis build whined:

> /Users/travis/build/mfontanini/libtins/src/packet_writer.cpp:78:5: warning: missing field 'comment' initializer [-Wmissing-field-initializers]

src/ip_reassembler.cpp
kyrofa: Of course-- done.

include/tins/ip_reassembler.h
kyrofa: Sure thing, no problem! When you see stuff like that either it's currently growing or something that's been deprecated; I made the wrong assumption :smiley: . Done.

src/packet_writer.cpp
mfontanini: Ah, that's true. Using that syntax is supposed to zero out the rest of the struct AFAIK, so I wonder why it's complaining. Anyway, that works as well, so it's fine.

kyrofa:
> Thanks for the fixes!

My pleasure, thanks for taking them, and thanks for the great lib! I was writing this stuff myself until I found libtins.

mfontanini:
Thank you!

PR #203, opened 2017-04-12
Add CMakeLists options LIBTINS_BUILD_EXAMPLES and LIBTINS_BUILD_TESTS

gpean:
This is to have control over the building of the examples and tests from e.g. parent CMake projects or from the command-line.

I left the default to still be building tests and examples by default.


mfontanini:
Thanks!

tthhcvv (approved):

Issue #226, opened 2017-07-18
Sniffer timeout and pcap_dispatch with libpcap 1.8


insomniacslk:
This is related to libpcap 1.8. I have noticed that with such version, a tool where I use sniff_loop and a timeout doesn't work anymore. This seems to be related to https://github.com/mfontanini/libtins/issues/41 . A working alternative (on my platform, Ubuntu 17.04 and Debian unstable) is to use pcap_dispatch instead of pcap_loop, as already suggested on the issue 41.

Subclassing seems a bit cumbersome, especially when it comes to keeping things in sync, so.. would it be possible to provide an alternative method that uses pcap_dispatch instead of pcap_loop? Or even better a flag that lets the caller suggest which method to use?

See BaseSniffer::next_packet, https://github.com/mfontanini/libtins/blob/master/src/sniffer.cpp#L215 

Thanks

mfontanini:
Interesting. I think letting the user choose what to use could be a good idea, although I wonder if this could just use `pcap_next` instead. The only difference I notice on the manpage (besides the timeout part) is that there's no mention of `pcap_breakloop` breaking the loop on `pcap_dispatch`, although in the "Return value" section they do mention that, which is weird.

Do you think it's safe to move this to use `pcap_dispatch`? I'm not sure why this didn't use this from the start (and I probably should have paid more attention to #41).

insomniacslk:
Yes, using `pcap_next` instead of `pcap_dispatch` would be equivalent, for how it is used in sniffer.cpp, as it is just a wrapper (see https://github.com/the-tcpdump-group/libpcap/blob/42c3865d71a3d3ad3fc61ee382ad3b5113d40552/pcap.c#L267). However you would need to set `oneshot_callback` on the pcap handle, and call the handler there (rather than letting `pcap_dispatch` do that for you). It would be probalby simpler to just go with `pcap_dispatch`.

If I had to choose I'd go for the least disruptive option, i.e. leave the default behaviour of `BaseSniffer::next_packet` unchanged, and add an optional argument to let the user switch between `pcap_loop` and `pcap_dispatch`. No existing code breaks, and we all win :)

On `pcap_breakloop`, I see that the man page of `pcap_dispatch` actually mentions it, see https://linux.die.net/man/3/pcap_dispatch :

```
pcap_dispatch() processes packets from a live
capture or ''savefile'' until cnt packets are processed,
the end of the current bufferful of packets is reached
when doing a live capture, the end of the ''savefile''
[...]

PR #228, opened 2017-07-21
Add SniffingMethod to specify either pcap_loop or pcap_dispatch

insomniacslk:
Allow the user to pass a suitable function pointer for the sniffing loop. In particular, the meaningful options would be `pcap_loop` and `pcap_dispatch`. The latter can be stopped with a timeout, the former doesn't (current behavior).

See https://github.com/mfontanini/libtins/issues/226 for background.

### DIFF
--- a/include/tins/sniffer.h
+++ b/include/tins/sniffer.h
@@ -45,6 +45,7 @@
 
 #include <pcap.h>
 
+
 namespace Tins {
 class SnifferIterator;
 class SnifferConfiguration;
@@ -72,7 +73,8 @@ public:
          * This constructor is available only in C++11.
          */
         BaseSniffer(BaseSniffer &&rhs) TINS_NOEXCEPT
-        : handle_(0), mask_(), extract_raw_(false) {
+        : handle_(0), mask_(), extract_raw_(false),
+          pcap_sniffing_method_(pcap_loop) {
             *this = std::move(rhs);
         }
 
@@ -85,6 +87,7 @@ public:
             swap(handle_, rhs.handle_);
             swap(mask_, rhs.mask_);
             swap(extract_raw_, rhs.extract_raw_);
+            swap(pcap_sniffing_method_, rhs.pcap_sniffing_method_);
             return* this;
         }
     #endif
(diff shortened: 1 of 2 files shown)


insomniacslk:
ouch, gcc 4.6.3 does not support C++11 type aliases :(

insomniacslk:
I can switch back to typedef, or the minimum compiler version could be bumped (gcc4.6 has incomplete C++11 support, 4.8 is generally good). Which one do you prefer?

include/tins/sniffer.h
```diff
@@ -45,6 +45,9 @@
 
 #include <pcap.h>
 
+using pcap_sniffing_method_t = int(*)(pcap_t*, int, pcap_handler, u_char*);
```
insomniacslk: Very good point. Any advice on the name here?

mfontanini: This should probably be inside `BaseSniffer` and something like `PcapSniffingMethod` should work. Also, I would actually do a hybrid of your initial version and this (using an enum). The reason is that we could have more methods (don't know if it's worth it) that way, also not being limiting with the interface that wants to be provided. So that way you don't even need to know which specific pcap function you want to grab the pointer from, but just look at the enum and use that.

insomniacslk: That's what I did before changing my mind again: carrying around a value from an `enum` and making sure that it's either `pcap_loop` or `pcap_dispatch`. Why did I change my mind and used a plain function ptr? Two reasons:
1) passing just a function ptr is simpler. It is also riskier, and requires to import `pcap.h`, but it is only necessary if you want to use something that is not the default `pcap_loop`
2) it is more flexible: I can pass a custom function that can wrap around `pcap_loop` or `pcap_dispatch`. Could be for unit testing, for implementing a different sniffing backend or whatever other reason, it's just more flexible. I personally like this approach and I have some idea for the future that would use this approach.

What do you think?

mfontanini: Fair enough, sounds like a good idea. This way we could actually write tests for `BaseSniffer` at least, which is pretty great. Can you just move the typedef inside `BaseSniffer` and rename it?

insomniacslk: Sure! I'll have access to my personal laptop soon and will update it

insomniacslk:
Updated as per the above feedback

insomniacslk:
Another question: if merged, would this change make it into a libtins release anytime soon? Basically this is the fix that would unblock my tool in debian/ubuntu (that run libpcap 1.8) so I am depending on it :)

mfontanini:
Looks good, I'll merge it later today or tomorrow morning at most. The next release will be major (version 4.0) so I want to break as many things as possible at once (e.g. this is an ABI breaking change as it adds a member to `BaseSniffer`) and there's still some things that need to be fixed before that, so the release won't be soon.

I've had some busy past few weeks but I'm going to have some extra time soon so I'll try to catch up on what's missing in the next couple of weeks.

PR #246, opened 2017-09-13
Consolidate 6 CMakeLists.txt under include/* into main CMakeLists.txt

solvingj:
This one is a very clean consolidation and simplification.


CMakeLists.txt
```diff
@@ -343,3 +342,13 @@ INSTALL(
     DESTINATION CMake
     COMPONENT dev
 )
+
+#Recursively list of all headers, with paths relative to include/tins
+FILE(GLOB_RECURSE INCLUDE_FILES RELATIVE include/tins "*.h")
```
mfontanini: Does the `RELATIVE` clause restrict the lookup to the `include/tins` directory or will this search for any `.h` files starting from the project's root?

solvingj: Indeed, it's project root, expression should now be "include/tins/*.h".

mfontanini: Yeah, that makes sense. Otherwise this will probably end up installing the googlemock headers globally.

solvingj: Have a look now, narrowed the scope and simplified.

mfontanini:
Thanks!

PR #247, opened 2017-09-14
Fixed compile failure somehow related to macro

solvingj:
Had three different machines here running Windows 10 with Visual Studio 2017 and 2015.  For current and clean master and develop branches with virtually every combination of options, compilation was failing out of the box with 800+ errors.  The exact cause was not clear, but Klemens made these two changes and compilation is now successful. Maybe from the change you can figure it out.  We could see that appveyor was able to compile, so it was a mystery what the difference was.  Hopefully, this change is trivial enough that you don't mind making it (assuming CI succeeds).

### DIFF
--- a/include/tins/ip_address.h
+++ b/include/tins/ip_address.h
@@ -202,9 +202,9 @@ private:
 
 #if TINS_IS_CXX11
 namespace std {
-
+template<typename T> struct hash; 
 template<>
-TINS_API struct hash<Tins::IPv4Address> {
+struct hash<Tins::IPv4Address> {
     size_t operator()(const Tins::IPv4Address& addr) const;
 };
 
--- a/include/tins/ipv6_address.h
+++ b/include/tins/ipv6_address.h
@@ -33,6 +33,7 @@
 #include <string>
 #include <iosfwd>
 #include <stdint.h>
+#include <functional> 
 #include "cxxstd.h"
 #include "macros.h"
 
@@ -220,9 +221,9 @@ private:
 
 #if TINS_IS_CXX11
 namespace std {
-
+template<typename T> struct hash; 
 template<>
-TINS_API struct hash<Tins::IPv6Address> {
+struct hash<Tins::IPv6Address> {
     size_t operator()(const Tins::IPv6Address& addr) const;
 };
 


mfontanini:
This seems to not build on clang (travis build). Also, are you sure this will work when building a dll? Removing the `TINS_API` macro (that's `__declspec(dllimport/export)`) should make that break once you try to use `hash<IPv4Address>` AFAIK (although I'm no Windows expert).

include/tins/cxxstd.h
```diff
@@ -31,6 +31,7 @@
 #define TINS_CXXSTD_H
 
 #include "config.h"
+#include <functional>
```
mfontanini: Why is this here? Isn't this trying to fix something on `ip_address.h`?

solvingj: We believe we've found a report of a bug with clang that explains the error:
https://github.com/martinmoene/gsl-lite/issues/63

Originally, `functional` was included in `cxxstd.h` because `cxxstd.h` was included in the others.  Commit b0013d2 now attempts to use workaround in the issue above by putting `#include <functional>` at the end of `ipaddress.h` and `ipv6_address.h`.

solvingj:
Regarding removing the TINS_API macro:  Yes, it was not complete the way it was, we've now moved the definition of the hash method to the header. This way the TINS_API method is no longer needed and the problem is avoided.

include/tins/ip_address.h
```diff
@@ -204,8 +205,11 @@ class TINS_API IPv4Address {
 namespace std {
 
 template<>
-TINS_API struct hash<Tins::IPv4Address> {
-    size_t operator()(const Tins::IPv4Address& addr) const;
+struct hash<Tins::IPv4Address> {
+    size_t operator()(const Tins::IPv4Address& addr) const
+    {
+        return std::hash<string>()(addr.to_string());
```
mfontanini: This used to perform the hash of the `uint32_t` value, can you roll it back so hashing an address is allocation free? (note to self, the same should be done for `IPv6Address`). Also, minor syntax suggestion, can you move the brace here so it looks like:

```) const { // < that one```

mfontanini:
Thanks!

PR #252, opened 2017-10-04
Exposing More Data Field in 802.11 Header

juno0812:
Added getter, setter, and tests for the more data field of the 802.11 header.

### DIFF
--- a/src/dot11/dot11_base.cpp
+++ b/src/dot11/dot11_base.cpp
@@ -164,6 +164,10 @@ void Dot11::power_mgmt(small_uint<1> new_value) {
     header_.control.power_mgmt = (new_value)? 1 : 0;
 }
 
+void Dot11::more_data(small_uint<1> new_value) {
+    header_.control.more_data = (new_value)? 1 : 0;
+}
+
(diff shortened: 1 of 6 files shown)


src/dot11/dot11_base.cpp
```diff
@@ -164,6 +164,10 @@ void Dot11::power_mgmt(small_uint<1> new_value) {
     header_.control.power_mgmt = (new_value)? 1 : 0;
 }
 
+void Dot11::more_data(small_uint<1> new_value) {
+    header_.control.more_data = (new_value)? 1 : 0;
```
mfontanini: You should be able to just do `= new_value` as that is already in the range [0, 1].

mfontanini:
Thanks, I'm not sure how this one was missed.

src/dot11/dot11_base.cpp
juno0812: I agree.  I was just keeping it consistent with the other fields like power management above.

mfontanini: Oh I didn't realize that was the case. Alright, I'll merge it as it is then.

PR #253, opened 2017-10-09
Add RawPDU c'tor for const payload_type &.

ajcollins:
Hello,

Whilst exploring libtins, the lack of a c'tor for RawPDU with payload_type when not in C++11 mode caught me out.

Small PR to add this c'tor, along the same lines as the C++11 move c'tor.


tests/src/raw_pdu_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include <tins/rawpdu.h>
+
+using namespace Tins;
+
+class RawPDUTest : public testing::Test {
+public:
+};
+
+TEST_F(RawPDUTest, ConstructFromPayloadType) {
+RawPDU::payload_type payload;
```
mfontanini: Can you indent this whole section 4 spaces to the right?

ajcollins: Of course - in fact, that was shocking that I'd not. Sorry!

PR #256, opened 2017-10-11
Fix header install location.

ajcollins:
Hi Matias,

When running `make install` on both Linux and macOS systems, the headers were being installed as `/usr/local/*.h` rather than the expected `/usr/local/include/tins/.*`.

Small PR to adjust the install location so that all headers and structure are copied as expected.

Best wishes,

Alex

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,11 +343,10 @@ INSTALL(
     COMPONENT dev
 )
 
-#Recursively list of all headers, with paths relative to include/tins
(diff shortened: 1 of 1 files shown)


mfontanini:
Oh god, this is so bad. There were some changes done in CMakeLists recently and I definitely tested that the installation worked correctly but then other changes happened and I probably didn't do it again. I'm so sorry, this is really bad.

Thanks for the fix!

ajcollins:
Not at all - thank you for the library and all of your hard work that has gone into it.

PR #257, opened 2017-10-12
Implemented matching for ND protocol related ICMPv6 messages

steven-chien:
I was trying to implement IPv6 next hop determination since some functions in my application require that I deliver packets with Ethernet frames attached to PacketSender and I needed next-hop link-layer info for the construction. I noticed that send_recv() does not correctly match ICMPv6 ND protocol responses since the checks are not implemented. I added the check of the flags and check if code equals zero according to https://tools.ietf.org/html/rfc4861#page-39 to ICMPv6 response matching.


mfontanini:
Thanks!

PR #260, opened 2017-10-24
Reorders SOURCE includes alphanumerically, and adds HEADERS as source

ajcollins:
Matias,

Whilst exploring the source in CLion, I was getting warnings about code completion not being available as header files did not form part of any target.

This PR adds targets with the headers, following the approach taken for the existing SOURCES. I've also reordered a couple to fit with the alpha sorting generally in use.

What do you think? Would you be happy to consider? Can easily review the repeated `${LIBTINS_INCLUDE_DIR}tins/` which could be stored in another variable, if desired?

Best wishes,

Alex


src/CMakeLists.txt
```diff
@@ -83,6 +83,93 @@ set(SOURCES
     utils/pdu_utils.cpp
 )
 
+set(HEADERS
+    ${LIBTINS_INCLUDE_DIR}tins/address_range.h
```
mfontanini: Can you add a slash before `tins` on all these paths? e.g. `${LIBTINS_INCLUDE_DIR}/tins/address_range.h`.

src/CMakeLists.txt
```diff
@@ -83,6 +83,93 @@ set(SOURCES
...
+    ${LIBTINS_INCLUDE_DIR}tins/detail/sequence_number_helpers.h
+    ${LIBTINS_INCLUDE_DIR}tins/detail/smart_ptr.h
+    ${LIBTINS_INCLUDE_DIR}tins/detail/type_traits.h
+    ${LIBTINS_INCLUDE_DIR}tins/dhcp.h
+    ${LIBTINS_INCLUDE_DIR}tins/dhcpv6.h
+    ${LIBTINS_INCLUDE_DIR}tins/dns.h
+    ${LIBTINS_INCLUDE_DIR}tins/dot3.h
+    ${LIBTINS_INCLUDE_DIR}tins/dot1q.h
+    ${LIBTINS_INCLUDE_DIR}tins/dot11.h
+    ${LIBTINS_INCLUDE_DIR}tins/dot11/dot11_base.h
```
mfontanini: I think dot11 related files should be split somewhere else as the whole protocol can be disabled so I don't think you should have it show up in your IDE if you explicitly disabled it.

mfontanini:
Oh this is pretty cool. Now I realize that when I was building this with VC (I don't normally use Windows) I noticed it was only showing source files when I opened the solution in it. I'll look into this later in more detail but it looks good besides a couple of comments I made!

src/CMakeLists.txt
ajcollins: Sure. I think that ends up with a double slash when concatentated with the variable but suspect CMake is agnostic to that.

src/CMakeLists.txt
ajcollins: Hadn’t realised that - sounds sensible - will sort that.

ajcollins:
Thanks - will make the changes later and update the PR.

ajcollins:
Matias,

I've changed the points in your comments - see what you think. I've followed the approach as per the libpcap dependent bits for consistency.

I've checked build and test with a `cmake ../ -DLIBTINS_ENABLE_DOT11=0` build. This seems fine.

I've not applied this to the tests. That'd probably require a change to the `CREATE_TEST` macro to pass the DOT11 headers. With CMake 3.1 `target_sources` could be used to add them, but the projects specifies 2.8.1. If desired, it may be more sensible to have a dummy target for those headers.

Best wishes,

Alex

mfontanini:
Looks good, thanks! For now tests don't matter too much. There's only a few headers in well and I don't think too many people want to change them anyway.

PR #264, opened 2017-10-28
Fix #263 - Install only the headers which are enabled.

### LINKED ISSUE #263: Disabled headers (e.g. dot11) are being installed anyway
If you disable dot11 via `-DLIBTINS_ENABLE_DOT11=0`, the headers for that protocol are installed regardless. This was introduced after ab2850e2 but that was a fix for another issue introduced before that so we can't just revert it.

ajcollins:
Hi Matias,

I saw you'd raised #263 - hadn't realised that was a requirement or the intended behaviour.

This PR uses the HEADERS list we'd defined previously to add the includes to the IDE, and installs those headers.

Best wishes,

Alex

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,12 +341,4 @@ INSTALL(
     EXPORT libtinsTargets
     DESTINATION CMake
     COMPONENT dev
-)
-
-# Install all headers in include/
-INSTALL(
-    DIRECTORY ${CMAKE_SOURCE_DIR}/include/
-    DESTINATION include
-    COMPONENT Headers
-    FILES_MATCHING PATTERN "*.h*"
-)
+)x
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -341,12 +341,4 @@ INSTALL(
...
-)
-
-# Install all headers in include/
-INSTALL(
-    DIRECTORY ${CMAKE_SOURCE_DIR}/include/
-    DESTINATION include
-    COMPONENT Headers
-    FILES_MATCHING PATTERN "*.h*"
-)
+)x
```
mfontanini: I think this `x` is breaking the travis/appveyor builds

mfontanini:
Thank you so much! You seem to have a typo but otherwise if it builds correctly I'll merge it.

ajcollins:
That was a clumsy typo. Must have knocked the key when committing. Corrected.

mfontanini:
Thanks, that works great!

PR #275, opened 2018-01-17
add  "missing" (bit) operator for hw_address, ip_address, ipv6_address

stubbfel:
Hi,
last i play a bit with ipv6 and address caclulation  and so i added some operator functions the address classes. 

* add or-operator and a simlple unit test for hw_address, ip_address, ipv6_address
* add not-operator and a simlple unit test for hw_address, ip_address, ipv6_address
* add greater-then-operator and a simlple unit test for ipv6_address
* add new constructor and a simlple unit test for network_interface, which use a ipv6_address to find the nic
* add override the function gateway_from_ip for ipv6_address parameter (untested)


stubbfel:
hmm looks like that some ipv6 adresses more invalid on windows  then on linux.

mfontanini:
Sorry it took me a few days to reply. This looks good! I'll have a deeper look later today or tomorrow and I'll probably merge it.

mfontanini:
Ugh sorry it took me so long to get back at this. I initially implemented `operator<` so we could store these addresses inside `std::map/set` and `operator==`/`operator!=` so you could... well, check for equality or inequality which is a thing you often do. In this case if we're adding `operator>`, I think it makes sense to implement `operator<=` and `operator>=` so address types are fully comparable. These should just be implemented in terms of the other operators (e.g. `operator<=` is implemented in terms of `operator>`).

Can you add those ones? It should be a one liner for all of them.

stubbfel:
ok, add these operators fo hw, ip and ip6 address

mfontanini:
I'm curious why some operators are implemented as friends while the same operator in a different class is just a regular operator?

stubbfel:
??? sry im not sure what you mean with implemented as friends? if i see correctly all operators, which i added are public and did not use the friend "visiblity". may you can give an example please?

mfontanini:
Sure, for example [IPv6Address::operator~](https://github.com/mfontanini/libtins/pull/275/files#diff-0ec71187163267465611a08c2c1d5172R264) is a free funcion. It's defined inside `IPv6Address` but since it's defined with friend visibility it's actually a free function (defined [here](https://github.com/mfontanini/libtins/pull/275/files#diff-3099c779cbc020e76bff2090159af21eR163)).

Then the same member but for [HWAddress::operator~](https://github.com/mfontanini/libtins/pull/275/files#diff-e29a34dd12279716ff1dde7f94c6958aR339) that's actually a member function. Is there any reason why they're not both defined in the same way?

stubbfel:
dam yes,  i think i just copied  & paste the friend signature for " TINS_API friend std::ostream& operator<<(std::ostream& os, const IPv6Address& addr);" and forget to remove the friend keyword
:man_facepalming:

stubbfel:
was the refactoring of the operator so ok?

mfontanini:
Yep, looks good! Thanks for the PR!

PR #276, opened 2018-01-27
Don't include dot11.h in tins.h if it is not configured in the library

ebertland:
I build the library with LIBTINS_ENABLE_DOT11=0. The installation procedure removes tins/dot11.h and tins/dot11/*.h from list of headers to install. Therefore, tins.h must not try to include dot11.h in this configuration of the library.


ebertland:
Sorry, should have used develop branch.

PR #281, opened 2018-02-01
Ignore (possibly malformed) options after EOL

ecatmur:
It is invalid for options (other than EOL) to occur after EOL. However,
most network stacks will stop processing options at EOL, meaning that
any further options will be ignored, including if they make the options
field malformed. We should likewise stop processing at EOL as otherwise
we will reject packets that most network stacks accept.

Similarly, when writing out options use EOL as end padding, instead of
NOP.

Test.


mfontanini:
This makes sense. Thanks!

PR #287, opened 2018-03-27
Add parsing of well known IPv6 extension headers

laudrup:
Add classes for IPv6 extension headers defined in the IPv6 protocol
specification (RFC 2460) as well as functions for creating them from
the IPv6 class' ext_header type.

The currently known extension headers are Hop-By-Hop Option,
Destination Routing, Routing and Fragment.

### DIFF
--- a/src/ipv6.cpp
+++ b/src/ipv6.cpp
@@ -43,6 +43,7 @@
 #include <tins/memory_helpers.h>
 #include <tins/detail/pdu_helpers.h>
 
+using std::make_pair;
 using std::vector;
 
 using Tins::Memory::InputMemoryStream;
@@ -95,7 +96,7 @@ IPv6::routing_header IPv6::routing_header::from_extension_header(const ext_heade
     routing_header header;
     header.routing_type = stream.read<uint8_t>();
     header.segments_left = stream.read<uint8_t>();
-    header.data = std::vector<uint8_t>(stream.pointer(), stream.pointer() + stream.size());
+    header.data.assign(stream.pointer(), stream.pointer() + stream.size());
     return header;
 }
 
@@ -212,15 +213,14 @@ uint32_t IPv6::get_padding_size(const ext_header& header) {
     return padding == 0 ? 0 : (8 - padding);
 }
 
-std::vector<IPv6::header_option_type> IPv6::parse_header_options(const uint8_t* data, size_t size)
-{
+vector<IPv6::header_option_type> IPv6::parse_header_options(const uint8_t* data, size_t size) {
     Memory::InputMemoryStream stream(data, size);
-    std::vector<header_option_type> options;
+    vector<header_option_type> options;
 
     while (stream.size() > 0) {
         try {
             uint8_t option = stream.read<uint8_t>();
-            if (option == OptionType::PAD_1) {
+            if (option == PAD_1) {
                 continue;
             }
             uint8_t size = stream.read<uint8_t>();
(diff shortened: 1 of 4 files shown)


src/ipv6.cpp
```diff
@@ -169,6 +212,33 @@ uint32_t IPv6::get_padding_size(const ext_header& header) {
...
+            if (option == OptionType::PAD_1) {
+                continue;
+            }
+            uint8_t size = stream.read<uint8_t>();
+            if (size > stream.size()) {
+                throw invalid_ipv6_extension_header();
+            }
+            if (option != PAD_N) {
+                std::vector<uint8_t> value(stream.pointer(), stream.pointer() + size);
+                options.push_back(std::make_pair(option, value));
```
mfontanini: So this sucks because I'm trying to keep the codebase both C++03 and C++11 compliant. If this was C++11 you would do:

```c++
options.push_back(make_pair(option, move(value)));
```

So you avoid creating an unnecessary copy of the vector. So in this case you can either have some ugly `ifdef TINS_IS_CXX11` and use move, otherwise do what this is moving or you can just move this into a single expression:

```c++
options.push_back(make_pair(option, vector<uint8_t>(stream.pointer(), stream.pointer() + size)));
```

Possibly having to split that in multiple lines if it's too long.

src/ipv6.cpp
```diff
@@ -69,6 +69,49 @@ PDU::metadata IPv6::extract_metadata(const uint8_t *buffer, uint32_t total_sz) {
...
+
+IPv6::routing_header IPv6::routing_header::from_extension_header(const ext_header& hdr) {
+    if (TINS_UNLIKELY(hdr.option() != ROUTING)) {
+        throw invalid_ipv6_extension_header();
+    }
+    Memory::InputMemoryStream stream(hdr.data_ptr(), hdr.data_size());
+    routing_header header;
+    header.routing_type = stream.read<uint8_t>();
+    header.segments_left = stream.read<uint8_t>();
+    header.data = std::vector<uint8_t>(stream.pointer(), stream.pointer() + stream.size());
```
mfontanini: You can just do

```c++
header.data.assign(stream.pointer(), stream.pointer() + stream.size());
``

File: src/ipv6.cpp [hunk 3]
```diff
@@ -169,6 +212,33 @@ uint32_t IPv6::get_padding_size(const ext_header& header) {
     return padding == 0 ? 0 : (8 - padding);
 }
 
+std::vector<IPv6::header_option_type> IPv6::parse_header_options(const uint8_t* data, size_t size)
+{
```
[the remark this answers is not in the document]
laudrup: Sure. Tried to keep the same coding style as the existing code, but I'm not surprised if I've missed a few places.

src/ipv6.cpp
laudrup: Of course. Thanks for pointing this out.

src/ipv6.cpp
```diff
@@ -169,6 +212,33 @@ uint32_t IPv6::get_padding_size(const ext_header& header) {
     return padding == 0 ? 0 : (8 - padding);
 }
 
+std::vector<IPv6::header_option_type> IPv6::parse_header_options(const uint8_t* data, size_t size)
```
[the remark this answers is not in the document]
laudrup: Will do.

src/ipv6.cpp
laudrup: I thought a bit about that actually and was pretty sure the compiler could probably optimize the copy away, but I never tested that.

Anyway since this library has to support C++03, I much prefer your second option over ugly #ifdefs, so I'll go with that.

We spotted.

laudrup:
Concerning the protocol format specifics, I haven't really been able to find any pcap files or other ways to tests this with "real life data" so this has mostly but written from what is described in RFC 2460.

Since your library makes it incredibly easy to write pcap files, I have tried writing my test data to pcap files and made sure that Wireshark and my code agreed on how the packets should be parsed, but that's the only kind of testing I could really think of.

mfontanini:
Thanks, looks good!

PR #293, opened 2018-05-08
Ensure local include directory comes before system.

banburybill:
If building (say) libtins 4.0 on a system with 3.4 installed, you need
the libtins include files to come from the repository include, not the
system include directory. The OpenSSL and PCAP includes may be from
the system include, so we need to ensure the libtins include is the
first on the list - which means the last on the before list.


mfontanini:
Thanks!

PR #298, opened 2018-05-17
Process more 802.11 options

RangelReale:
- HT capability (45)
- VHT capability (191)
- EXT capability (127)

### DIFF
--- a/include/tins/dot11/dot11_base.h
+++ b/include/tins/dot11/dot11_base.h
@@ -127,9 +127,12 @@ public:
         ERP_INFORMATION,
         TS_DELAY,
         TCLAS_PROCESSING,
-        QOS_CAPABILITY = 46,
+        HT_CAPABILITY = 45,
+        QOS_CAPABILITY,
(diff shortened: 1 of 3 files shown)


include/tins/sniffer.h
```diff
@@ -673,7 +673,7 @@ void Tins::BaseSniffer::sniff_loop(Functor function, uint32_t max_packets) {
     for(iterator it = begin(); it != end(); ++it) {
         try {
             // If the functor returns false, we're done
-            #if TINS_IS_CXX11 && !defined(_MSC_VER)
+			#if TINS_IS_CXX11 && defined(TINS_IS_CXX11_MSVCCOMPAT)
```
mfontanini: Nice, this was added a while back when VC's support for C++11 was still not great.

mfontanini: There's tabs here, could you use spaces?

RangelReale: Done!

PR #299, opened 2018-05-29
Reduction of memory allocations for hw_address_to_string

thrimbor:
During my usage of a profiler to optimize a program using libtins to analyze a pcap file and collect statistics, I noticed many temporary allocations for a string stream and traced it to hw_address_to_string.
I rewrote it to use a string instead, which results in only a single allocation per function call. The code isn't very pretty, but it did have a noticeable effect on the performance of my program.

I also removed storage_to_string since hw_address_to_string was its only user.

### DIFF
--- a/src/hw_address.cpp
+++ b/src/hw_address.cpp
@@ -34,32 +34,35 @@
 #include <tins/exceptions.h>
 
 using std::string;
-using std::ostream;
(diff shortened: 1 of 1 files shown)


mfontanini:
Nice, thanks!

PR #306, opened 2018-08-02
Added base class access specifier for socket_close_error

UlfWetzker:
I think the access specifier got lost during commit af71a4eca7b9d284ccca879118871c1b3f070cab.


mfontanini:
Oops, thanks!

PR #315, opened 2018-09-25
Add QinQ frame

teodoro-vargas:
Current implementation of libtins uses Dot1Q frame to represent both Dot1Q and Dot1AD (QinQ). The serialization of a packet with double tagging writes the PIDs in the wrong order: 0x8100 (Dot1Q) followed by 0x88a8 (Dot1AD), when the order should be 0x88a8 (Dot1AD) followed by 0x8100 (Dot1Q).


mfontanini:
Wouldn't it be simpler to simply fix `Dot1Q` instead of introducing a new class to fix this? e.g. flip the check and instead of doing what it's currently doing with the `inner_pdu`, do it with the `parent_pdu`.

teodoro-vargas:
@mfontanini That would require to assume that the parent frame is an `EthernetII`, since the method `parent_pdu` returns a `PDU` and a cast would be necessary. We can try that option, but we think that that casting can be a little dangerous.

mfontanini:
You don't have to cast the parent PDU, you can just call `PDU::pdu_type` and see if it's the ethernet one.

teodoro-vargas:
Ok, that didn't work for me. However, keeping the change in the `write_serialization` method in `EthernetII` and removing the Dot1AD frame did. 
Still, having to get the `Dot1Q` frame and then check if there's another `Dot1Q` frame inside, instead of just looking for the `Dot1AD` frame, would be a little annoyance for some people.

teodoro-vargas:
@mfontanini I believe that the approach you mentioned didn't work because of the `pdu_type` value during the serialization process (which is a static value of 0x8100, in Dot1Q). In the latest commit of this PR, It's checked directly in the EthernetII frame if the following two frames are Dot1Q, in which case, a QinQ PID is written (0x88a8).
Please, let us know if the change is still incorrect.

teodoro-vargas:
@mfontanini friendly ping. We applied the fixes you requested. 
Thanks.

mfontanini:
Looks good! Do you mind adding a test? Specifically something that constructs a simple packet like `EthernetII() / Dot1Q() / Dot1Q() / something` using some tag for each dot1q frame, the serializing it, parsing it back and making sure everything works?

PR #318, opened 2018-11-02
windows build: disable integer conversion warnings

douglaspeer:


mfontanini:
Are there normally a lot of these? At some point I went through all of them and fixed them but some more could have come up since then. I think I'd rather fix them instead of snoozing the warnings.

douglaspeer:
I agree with your assessment.  There are 286 warnings if I recall correctly.  I am running msvc 2017.  Most of them were size_t to uint32_t warnings.  I wasn't sure if the desired outcome of squashing these was to update the API to use size_t or perform static_cast<uint32_t>, or even a mixture of the two.

I started by modifying the API, but in many cases it felt like overkill, or there was already a mixture of 64bit -> 32bit squashing.

Please advise.

mfontanini:
Woah that's a lot of warning. The API shouldn't be changed, the types should be converted instead.

douglaspeer:
I'll dig into this at some point, and provide a new pull request.

PR #319, opened 2018-12-01
Fix Dot1Q serialization for non-qinq packets

saxonparker:
The recent change to serialize QinQ packets properly broke serialization of single VLAN tagged packets. The serialize code in `EthernetII` was not setting the payload type of Dot1Q packets. This PR is a simple fix for that.


mfontanini:
Thanks!

PR #320, opened 2018-12-13
Some buffer overflow related bug fixes for radiotap.

DDoSolitary:


src/utils/radiotap_parser.cpp
```diff
@@ -186,6 +186,9 @@ RadioTap::option RadioTapParser::current_option() {
 }
 
 const uint8_t* RadioTapParser::current_option_ptr() const {
+    if (TINS_UNLIKELY(current_ptr_ > end_)) {
```
mfontanini: Given that `end_` is already out of bounds, this should probably be checking for `>= end_`, right?

DDoSolitary: But using `>=` will lead to test failures in `RadioTapWriter`.

mfontanini: I see, I don't think the way the writer is using this is wrong. I don't think we should add this check here, as it's not covering all cases as it is. The documentation should probably be more explicit and say something like "if the parser has no fields left, dereferencing the returned pointer is undefined behavior", as this seems more of a safe check in case the user mis-uses the class (e.g. by not calling `has_fields` before calling this).

DDoSolitary: But then passing in a malformed packet like `RadioTapTest::expected_packet7` would be some sort of undefined behavior. `malformed_packet()` will only be thrown in `align_buffer()` when aligning is required for the fields that overflows the buffer.

So think we should either: 

- Remove the overflow check in `align_buffer()` and `current_option_ptr()`.
- Use `EXPECT_FALSE(has_fields())` instead of `EXPECT_THROW` in the `ConstructorFromBuffer6` test.
- Clarify in documentation that users should perform safe check themselves.

Or:

Keep the checks that I proposed.

DDoSolitary: BTW, I personally prefer the latter coz we already have the check in `current_option()`.

mfontanini: I don't understand why you're proposing to remove the check in `align_buffer`. What would you gain from doing this?

My idea with parsing radiotap was to make it fast and lazy, hence things throwing when taking data out of the radiotap class rather than right during parsing. You're not going to pay for radiotap parsing unless you want to (and a lot of times you don't care about radiotap at all). 

`current_option` is very different from `current_option_ptr`. You can get a past-the-end pointer and still do something useful with it (see the radiotap writer), whereas if you didn't do this check in `current_option`, you could return bogus data which really serves no purpose at all.

As it is, the check you added doesn't check all of the potential "invalid cases" and therefore I don't think it makes sense for it to be there. If I call this when `current_ptr_ == end_ + 1`, this throws but if they're the same it won't throw even though the returned pointer is just as invalid as the other one.

DDoSolitary: > I don't understand why you're proposing to remove the check in `align_buffer`. What would you gain from doing this?

This check will make some malformed packet to throw while some not.

DDoSolitary: For example, consider that we pass the following option buffers to `RadioTapParser`'s constructor, and then call `advance_field()`:

---

```
03 00 00 00 00
```

`it_present` specifies Flags(u8) and Rate(u8), but only Flags is present.

```

```
0A 00 00 00 00
```

`it_present` specifies Flags(u8) and Channel(u16+u16), but only Flags is present.

---

Rate in the former buffer requires no alignment so `advance_field()` won't throw, but Channel in the latter requires a 4-byte alignment so `align_buffer()` will throw `malformed_packet()` However, from users' perspective both buffers are simply illegal data. I think such inconsistency will make error handling much more frustrating.

If we delete the checking in `align_buffer()`, it's OK for the function itself coz it doesn't read or write the memory that the pointer points to, and users can still perform safety checking via `has_fields()`.

PR #322, opened 2018-12-25
Buffer overflow fixes for radiotap.

DDoSolitary:
Previous discussions: #320

### DIFF
--- a/include/tins/utils/radiotap_parser.h
+++ b/include/tins/utils/radiotap_parser.h
@@ -114,7 +114,11 @@ public:
     RadioTap::option current_option();
 
     /**
-     * Gets the pointer at which the current option is located
+     * \brief Gets the pointer at which the current option is located
+     *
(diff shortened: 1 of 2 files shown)


mfontanini:
Sorry it took me a while to merge this. Thanks for the PR!

DDoSolitary:
I just tried to build v4.2 in a x86 QEMU machine and the test for radiotap failed. It was the same failure in the same environment that let me notice the bug and prepare the PR. Actually, I'm quite surprised to see this because all the tests has passed before I submitted this PR months ago. I think I must have made some mistakes when testing the patch, but anyway, it's clear that the bug hasn't been completely fixed yet. I'm still investigating this and will try to prepare another patch to fix the failure I'm experiencing now.

DDoSolitary:
@mfontanini  It turns out that the failures are caused by float precision problems. I also fixed some other bugs, and the PR is ready: #333

Issue #300, opened 2018-06-06
the Frame Length does not get written correctly


pepper-jk:
When looking at packets in .pcap files there are two lengths defined in a header:
* the **Frame Length**, which provides the length of the packet, which was actually send in real world
* the **Capture Length**, which provides the length of the packet, which is stored in the pcap

![packet_header](https://user-images.githubusercontent.com/19859589/41036766-6bc7c734-6991-11e8-87cb-29754281164a.png)

This allows the capture to be a lot smaller then the real traffic that was monitored, while still providing the information about how much data was transferred. This **information gets lost** when using Libtins to read a .pcap and write it out again.

This is due to the **[Packet](https://github.com/mfontanini/libtins/blob/a6817528bc237970bca4d0351195151552f01db6/include/tins/packet.h#L315) data structure**, which only stores the [PDUs content](https://github.com/mfontanini/libtins/blob/a6817528bc237970bca4d0351195151552f01db6/include/tins/pdu.h) and its timestamp.

As far as I looked into it the sniffer does not read the information either since there is no data structure to store it in.

As you can see the **Frame Length** and **Capture Length** are set to the **size of the packet**, when writing:
https://github.com/mfontanini/libtins/blob/a6817528bc237970bca4d0351195151552f01db6/src/packet_writer.cpp#L72-L80

I'm not sure if you consider this a bug, or if you do not want handle this type of "metadata".

This was noticed when using the [ID2T-toolkit](https://git.tk.informatik.tu-darmstadt.de/SPIN/ID2T-toolkit) to inject attack packets into a .pcap file. ID2T uses libtins for reading and writing pcaps, while not modifying the original packets, but only adding new packets on top.

Before:

    $ capinfos resources/201804041400_200000PDUs.pcap 
    File name:           resources/201804041400_200000PDUs.pcap
    File type:           Wireshark/... - pcapng
    File encapsulation:  Ethernet
    File timestamp precision:  microseconds (6)
    Packet size limit:   file hdr: (not set)
    Packet size limit:   inferred: 34 bytes - 96 bytes (range)
    Number of packets:   200 k
    File size:           17 MB
    Data size:           125 MB
    Capture duration:    2.482684 seconds
    First packet time:   2018-04-04 07:00:00.448274
    Last packet time:    2018-04-04 07:00:02.930958
    Data byte rate:      50 MBps
    Data bit rate:       404 Mbps
    Average packet size: 627.71 bytes
    Average packet rate: 80 kpackets/s
[...]

PR #324, opened 2019-01-16
Fix frame length

pepper-jk:
This pull request should fix issue #300.

I implemented a virtual method `PDU::advertised_size()`, which defaults to the sum of `header_size() + trailer_size()` plus `advertised_size()` of all inner PDUs down to the IP PDU.

The `IP::advertised_size()` method returns the `tot_len` instead.

The PacketWriter now uses this virtual method to determine the frame length before writing the packet out.

Please let me know if this solution is to your liking.

### DIFF
--- a/include/tins/ip.h
+++ b/include/tins/ip.h
@@ -281,6 +281,10 @@ public:
 
     /* Getters */
 
+    uint32_t advertised_size() const {
+        return static_cast<uint32_t>(tot_len());
+    }
+
     /**
      * \brief Getter for the header length field.
      *
--- a/src/pdu.cpp
+++ b/src/pdu.cpp
@@ -85,6 +85,14 @@ uint32_t PDU::size() const {
     return sz;
 }
 
+uint32_t PDU::advertised_size() const {
+    uint32_t result = header_size() + trailer_size();
+    if (inner_pdu_) {
+        result += inner_pdu()->advertised_size();
+    }
+    return result;
+}
+
 void PDU::send(PacketSender &, const NetworkInterface &) { 
     
 }
--- a/include/tins/pdu.h
+++ b/include/tins/pdu.h
@@ -281,6 +281,12 @@ public:
      */
     uint32_t size() const;
 
+    /** \brief The whole chain of PDU's advertised size, including this one.
+     *
+     * Returns the sum of this and all children PDU's advertised size.
+     */
+    virtual uint32_t advertised_size() const;
+
     /**
      * \brief Getter for the inner PDU.
      * \return The current inner PDU. Might be a null pointer.
--- a/src/packet_writer.cpp
+++ b/src/packet_writer.cpp
@@ -70,12 +70,12 @@ void PacketWriter::write(Packet& packet) {
 }
 
 void PacketWriter::write(PDU& pdu, const struct timeval& tv) {
-    PDU::serialization_type buffer = pdu.serialize();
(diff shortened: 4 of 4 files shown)


src/pdu.cpp
```diff
@@ -85,6 +85,20 @@ uint32_t PDU::size() const {
     return sz;
 }
 
+uint32_t PDU::advertised_size() const {
+    uint32_t sz = header_size() + trailer_size();
+    const PDU* ptr(inner_pdu_);
+    while (ptr) {
+        sz += ptr->advertised_size();
+        if(ptr->pdu_type() != IP) {
```
mfontanini: This check shouldn't be here. This has to work no matter what PDUs a packet contains. I think a better and generic solution to this is to have the default implementation return `header_size() + trailer_size() + inner_pdu()->size()`  (this last one obviously only if there's an inner PDU). This way you're defining the advertised PDU size to be the size of the current PDU + whatever the next PDU tells you. So if the next PDU is IP, then it will simply return its total length without going further down the chain.

include/tins/ip.h
```diff
@@ -281,6 +281,11 @@ class TINS_API IP : public PDU {
 
     /* Getters */
 
+    uint32_t advertised_size() const {
+        uint32_t sz = static_cast<uint32_t>(tot_len());
```
mfontanini: Let's just make this a single return, no need for an intermediate variable.

pepper-jk: Got it.

src/pdu.cpp
mfontanini: Sorry, I mean `inner_pdu()->advertised_size()` here.

pepper-jk: Well the IP::tot_len() contains the lenght of all inner PDUs does it not?

That was my understanding after looking at the IP header in wireshark.

mfontanini: Exactly, which is why IP can get away with just returning `tot_len` as its advertised size vs the rest of the PDUs which need to keep going down the PDU chain to find out.

pepper-jk: I see the check is no longer needed, since we IP::advertised_size() will be the recursion anchor.

I'll fix it.

include/tins/ip.h
pepper-jk: Made the changes, just waiting for your continuous integration.

src/pdu.cpp
pepper-jk: Made the changes, just waiting for your continuous integration.

src/pdu.cpp
```diff
@@ -85,6 +85,10 @@ uint32_t PDU::size() const {
     return sz;
 }
 
+uint32_t PDU::advertised_size() const {
+    return header_size() + trailer_size() + inner_pdu()->advertised_size();
```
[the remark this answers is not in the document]
pepper-jk: My bad. Fixed.

pepper-jk:
Is there anything else that needs changing?

mfontanini:
Nope, it looks good. I just didn't get to merge this. Thanks!

pepper-jk:
Any updates on the patch version?

mfontanini:
Oops, I completely forgot. I'll definitely do this tonight! Sorry about that

mfontanini:
I just pushed v4.2. Sorry for the delay!

PR #330, opened 2019-02-07
Fix check whether we are crosscompiling

danvratil:
CMake defines CMAKE_CROSSCOMPILING when crosscompiling so
use that instead of the undefined CROSS_COMPILING variable.


mfontanini:
Thanks!

PR #331, opened 2019-02-20
cmake: update location of cmake configuration files for unix subsystems

Kaisrlik:
Default cmake configuration path is different for unix subsystems [1].

[1]: https://cmake.org/cmake/help/v3.8/command/find_package.html


mfontanini:
This still works as it is and it's just a recommendation but let's use a more default path. Thanks!

PR #333, opened 2019-03-10
Overhaul RadioTapWriter

DDoSolitary:


DDoSolitary:
Sure, though it won't be possible to add test case for ba2216e6e923c65798ccce988af485ffc7ac74d2, because it only causes problems on very specific environments like qemu-system-x86.

mfontanini:
Yeah I imagined that wouldn't normally happen. Mostly for the padding one, I'm surprised I didn't caught that with some existing test.

DDoSolitary:
@mfontanini Done.

mfontanini:
Thanks!

PR #334, opened 2019-03-13
simplify tcp flag checks, fix stream_follower

Karry:
On various places was used simple comparison for checking state of flags.

`tcp.flags() == (TCP::SYN | TCP::ACK)`

This is not what you want usually, because this check is false
in case that another flag is set also. Correct check for syn-ack
packet should be:

`(tcp.flags() & (TCP::SYN | TCP::ACK)) == (TCP::SYN | TCP::ACK)`

To simplify this kind of check, add new has_flags method:

`bool TCP::has_flags(small_uint<12> check_flags) const`

### DIFF
--- a/src/tcp_ip/stream_follower.cpp
+++ b/src/tcp_ip/stream_follower.cpp
@@ -84,7 +84,7 @@ void StreamFollower::process_packet(PDU& packet, const timestamp_type& ts) {
     if (iter == streams_.end()) {
         // Start tracking if they're either SYNs or they contain data (attach
         // to an already running flow).
-        if (tcp->flags() == TCP::SYN || (attach_to_flows_ && tcp->find_pdu<RawPDU>() != 0)) {
+        if (tcp->has_flags(TCP::SYN) || (attach_to_flows_ && tcp->find_pdu<RawPDU>() != 0)) {
             iter = streams_.insert(make_pair(identifier, Stream(packet, ts))).first;
             iter->second.setup_flows_callbacks();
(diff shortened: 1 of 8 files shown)


src/tcp_ip/stream_follower.cpp
```diff
@@ -84,7 +84,7 @@ void StreamFollower::process_packet(PDU& packet, const timestamp_type& ts) {
     if (iter == streams_.end()) {
         // Start tracking if they're either SYNs or they contain data (attach
         // to an already running flow).
-        if (tcp->flags() == TCP::SYN || (attach_to_flows_ && tcp->find_pdu<RawPDU>() != 0)) {
+        if (tcp->has_flags(TCP::SYN) || (attach_to_flows_ && tcp->find_pdu<RawPDU>() != 0)) {
```
mfontanini: Minor but I'd probably create a variable like

```c++
const bool is_syn = tcp->has_flags(TCP::SYN);
```

And then use it in both conditions below.

mfontanini:
Nice, good catch!

src/tcp_ip/stream_follower.cpp
Karry: It is up to you, if you think that code will be more readable. Or you are worry about performance?

mfontanini: Not really performance related, just so we avoid having the same expression twice.

mfontanini:
Thanks!

PR #335, opened 2019-03-18
Fix detection of a new TCP flow

martin-beran:
If not attaching to existing flows, a new TCP flow starts by the first packet from the client, which has SYN without ACK, and not by the reply packet from the server, which has SYN+ACK.


mfontanini:
Thanks!

PR #340, opened 2019-04-08
Bugfix/crosscompilation does not properly set include dir

Kaisrlik:
After cross-compilation I observe the variable `LIBTINS_INCLUDE_DIRS` in `libtinsConfig.cmake` file contains wrong value. For the reason I slightly rework it and make the include directory part of the target

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,6 +347,8 @@ INSTALL(
     COMPONENT dev
 )
 
+TARGET_INCLUDE_DIRECTORIES(tins INTERFACE "$<INSTALL_INTERFACE:include/>")
+
 # Install the export set for use with the install-tree
 INSTALL(
     EXPORT libtinsTargets


cmake/libtinsConfig.cmake
```diff
@@ -1,16 +1,9 @@
 # - Config file for the libtins package
-# It defines the following variables
-#  LIBTINS_INCLUDE_DIRS - include directories for libtins
-#  LIBTINS_LIBRARIES    - libraries to link against
```
mfontanini: Is there a way not to remove these? If people are currently relying on these being available, it's going to break their code.

Kaisrlik: Honestly, I think this change will force all users to work with your library/CMake in correct way. 

What I tested so far I was not able produce the problem when it can break the build.
If I'm right It can break the build only in case when they are using the library in "header-only" mode. Is it possible to use the library in that mode?
If you know any problematic case feel free to share it with me.

mfontanini: There's not really a "header only" mode. Going back to this change: we can't introduce a breaking change in the CMake script and just say "deal with it". Whatever code is out there that's using this file should still work after this change.

Kaisrlik: I've just tried to convince you ;]  Hopefully I will find a time to update it this week.

Kaisrlik:
Hello Matias,
I played with cmake and I cannot find a way how to fix `LIBTINS_INCLUDE_DIRS` to work in all cases. Based on the cmake configuration described in [1] current configuration is alright.
In addition I've browsed different projects to see how are using it. All projects what I saw they were using it in way of "targets" - similar to the first commit what I sent.
For the reason I've kept the setting and added `target_include_directory`.
[1]: https://gitlab.kitware.com/cmake/community/wikis/doc/tutorials/How-to-create-a-ProjectConfig.cmake-file

Issue #339, opened 2019-04-08
Memory leak in active tests


m-peko:
Hi there,

Currently, I am making some performance tests in this project and, while doing so, I ran into following lines of code in `libtins/tests/active_tests/include/active_test_runner.h`:

```
template <typename T>
void ActiveTestRunner::add_test() {
    tests_.emplace_back(new T(packet_sender_, configuration_));
}
```

I think they are not safe because they would create memory leak. The [documentation](https://en.cppreference.com/w/cpp/container/vector/emplace_back) says that if an exception is thrown, the call to emplace has no effect - which means the plain pointer we have passed in is never deleted.

Therefore, I would suggest changing lines above to something like this:

```
template <typename T>
void ActiveTestRunner::add_test() {
    tests_.emplace_back(std::unique_ptr<T>(new T(packet_sender_, configuration_)));
}
```

What do you think? [PR](https://github.com/mfontanini/libtins/pull/341) is made with this fix.

mfontanini:
Comments in the PR. Thanks!

PR #341, opened 2019-04-08
Fix possible memory leak in active tests

m-peko:
Fix for issue #339


mfontanini:
While in theory this is an issue, in practice it isn't. The only way for this to have a leak is if `emplace_back` throws and the only way it would throw in this scenario is if you ran out of memory and at that point, you have worse issues. Anyway, I'll merge it but I don't think you'd ever hit this condition in practice.

PR #343, opened 2019-04-11
Fix several leaks

pallas:
This series fixes two issues.

1. Using pcap_geterr after handle_ is closed.
2. Failing to free filter_ and handle_ on construction error.

Found by Valgrind.


mfontanini:
Ugh this sucks. Maybe some day the library will be C++11 only and will be able to use `unique_ptr` for these type of things.

Thanks!

PR #346, opened 2019-04-24
radiotap: avoid unused variable warning

pallas:


src/utils/radiotap_parser.cpp
```diff
@@ -132,6 +132,7 @@ struct RadioTapFlags {
 #endif
 
 void align_buffer(const uint8_t* buffer_start, const uint8_t*& buffer, uint32_t size, size_t n) {
+    (void)size;
```
mfontanini: This is a local function so I'd rather just remove the `size` parameter rather than ignoring it. It's only called once in this file so that should be straightforward.

pallas:
done

mfontanini:
Thanks!

Issue #360, opened 2019-07-28
C2664 Error on sniffer.h


realsarm:
Hi there

The following is my code that is actully the http example of libtins.
I use Visual Studio 2017 v141

`#define WIN32
#define TINS_STATIC
#define NOMINMAX
#pragma warning(disable : 4996)
#include <string>
#include <iostream>
#include <stdexcept>
#include <boost/regex.hpp>
#include <tins/tcp_ip/stream_follower.h>
#include <tins/sniffer.h>
#include <tins/tins.h>
#include "color.h"
#include <vector>
#include <sstream>

using std::string;
using std::cout;
using std::cerr;
using std::endl;
using std::exception;
using std::vector;
using std::wcout;
using std::stringstream;

using boost::regex;
using boost::match_results;

using Tins::Packet;
using Tins::Sniffer;
using Tins::SnifferConfiguration;
using Tins::TCPIP::Stream;
using Tins::TCPIP::StreamFollower;
using Tins::NetworkInterface;
using termcolor::on_red;
using termcolor::on_green;
using termcolor::reset;

// This example captures and follows TCP streams seen on port 80. It will
// wait until both the client and server send data and then apply a regex
// to both payloads, extrating some information and printing it.

// Don't buffer more than 3kb of data in either request/response
const size_t MAX_PAYLOAD = 3 * 1024;
// The regex to be applied on the request. This will extract the HTTP
// method being used, the request's path and the Host header value.
regex request_regex("([\\w]+) ([^ ]+).+\r\nHost: ([\\d\\w\\.-]+)\r\n");
// The regex to be applied on the response. This finds the response code.
regex response_regex("HTTP/[^ ]+ ([\\d]+)");

void on_server_data(Stream& stream) {
	match_results<Stream::payload_type::const_iterator> client_match;
	match_results<Stream::payload_type::const_iterator> server_match;
	const Stream::payload_type& client_payload = stream.client_payload();
	const Stream::payload_type& server_payload = stream.server_payload();
	// Run the regexes on client/server payloads
	bool valid = regex_search(server_payload.begin(), server_payload.end(),
		server_match, response_regex) &&
		regex_search(client_payload.begin(), client_payload.end(),
			client_match, request_regex);

	stringstream ss;
	for (char c : server_payload) {
		ss << c;
	}

	cout << on_green << "Server raw payload " <<
		ss.str() << reset << endl;

	// If we matched both the client and the server regexes
	if (valid) {
		// Extract all fields
[...]

PR #362, opened 2019-07-29
Resolves C2696 error while using VS 2017 and up

realsarm:
This is for issue #360


mfontanini:
`PDU` isn't a `Packet`, this isn't a safe cast. This should be checking the VS version instead. Some older version didn't have enough C++11 support to handle the template types used to do this PDU/Packet dispatching.

PR #369, opened 2019-10-01
add IPv6 check for Link-Local unicast address

Karry:
Hi. In our project, it is useful to check if some IPv6 address is link-local. What about to add this check to libtins directly?


thom1337:
+1

mfontanini:
Thanks!

PR #374, opened 2019-10-19
Fix malformed packet exception on esp header

### LINKED ISSUE #373: Malformed packet exception on ipv6 with esp header
When ipv6 packet has esp header libtins throw malformed packet exception. Example pcap is attached to issue.

As i can see, a problem is that esp extension does not have ordinary extension header format. See [rfc2406](https://tools.ietf.org/html/rfc2406#page-3). It just has SPI and Sequence number and that is all, trailing bytes are encrypted and libtins' ipv6 parser cannot does much about it. Better just pass it to ipsec parser that can parse SPI, sequence number and trailing bytes will be RawPDU.

I'l try to do Pull Request

[esp.pcap.zip](https://github.com/mfontanini/libtins/files/3746905/esp.pcap.zip)

almikhayl:
Fix #373


mfontanini:
Thanks and sorry for the delay!

PR #382, opened 2019-12-20
Don't include non-existing headers when installed without libpcap

laudrup:
If libpcap support has been disabled (LIBTINS_ENABLE_PCAP=OFF)
then the headers requiring that library will not be installed,
but they will still be included from the main tins.h convenience
header.

This fixes that by sorrounding the includes with an #ifdef the
same way it has been done for DOT11 support.


mfontanini:
Thanks!

PR #383, opened 2019-12-20
Fix compiler warnings from MSVC 2017

laudrup:
The MS C++ compiler doesn't seem to understand that a small constant
like 0, although strictly being an integer, will fit perfectly fine
into an 8 bit type so add some safe casts to silence that warning.


mfontanini:
Thanks!

PR #402, opened 2020-06-30
generate config.h in binary directory

lzy1g1225:
I found that libtins got recompile everytime which drive me crazy when I switched debug/release mode or when I switched from win32/linux. 
The reason is the cmake generates config.h for those build types and got the config.h overwritten, which causing the compiler to recompile the whole project.
I suggest to generate config.h in binary directory. That make every config.h stay with there own build type.


lzy1g1225:
Looks like the appveyor build check failed because tests and examples didn't update include path for config.h. Will add this later.

PR #426, opened 2020-12-29
Added RFC8335 Extended echo types to headers

theDogOfPavlov:
Trivial commit to add the RFC8335 Extended Echo request/reply types to the headers for both ICMPv4 and ICMPv6.


theDogOfPavlov:
looks like a build tools error, rather than anything to do with the commit...

mfontanini (approved):

mfontanini:
Thanks!

theDogOfPavlov:
my pleasure!

PR #437, opened 2021-03-24
Fix wrong address endian

ceerRep:
Host endian has been implicitly converted to big endian in "IPv4Address::operator uint32_t()"


mfontanini:
Nice, thanks!

PR #439, opened 2021-04-23
Fix getter being labeled as setter in dns header file comments

adriancostin6:
While reading the DNS class reference in the documentation I noticed that the getter methods were labeled as setters in the description field, so I fixed the comments in the header file to solve that.


mfontanini:
Nice catch, thanks!

PR #449, opened 2021-07-30
SNAP: allow the eth_type to be explicitly set

fhuberts:
Otherwise it is overwritten by write_serialization


fhuberts:
I'm a new user of the library and it might be that I'm doing something wrong, but I did not see a way to set the SNAP eth_type to something explicitly.

PR #453, opened 2021-08-12
All: fix warnings when compiling with -Wzero-as-null-pointer-constant

fhuberts:

### DIFF
--- a/include/tins/hw_address.h
+++ b/include/tins/hw_address.h
@@ -131,7 +131,7 @@ public:
      * 
      * \param ptr The pointer from which to construct this address.
      */
-    HWAddress(const storage_type* ptr = 0) {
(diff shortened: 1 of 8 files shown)


include/tins/hw_address.h
```diff
@@ -131,7 +131,7 @@ class HWAddress {
      * 
      * \param ptr The pointer from which to construct this address.
      */
-    HWAddress(const storage_type* ptr = 0) {
+    HWAddress(const storage_type* ptr = nullptr) {
```
mfontanini, line 134: This breaks in C++ < 11. We probably should drop support for C++03 and enforce people to use C++11 but until then these sort of changes can't be made or must be guarded

fhuberts: Right, no problem.
I'll just carry the patch in our repo.

PR #460, opened 2021-11-19
Fix a great number of warnings

fhuberts:


fhuberts:
I've squashed all warning fixes into 1 commit, it became too unwielding to carry each commit.

PR #466, opened 2022-02-19
dns: bad label size interpreted as decompression pointer

bwillcox:


src/dns.cpp
```diff
@@ -342,7 +342,7 @@ uint32_t DNS::compose_name(const uint8_t* ptr, char* out_ptr) const {
             throw dns_decompression_pointer_loops();
         }
         // It's an offset
-        if ((*ptr & 0xc0)) {
+        if (((*ptr & 0xc0) == 0xc0)) {
```
mfontanini, line 345: Would it make more sense to check for `*ptr == 0xc0`?

tests/src/dns_test.cpp
```diff
@@ -551,3 +551,54 @@ TEST_F(DNSTest, SOARecordSerialize) {
     EXPECT_EQ(0x8ad71928U, r2.expire());
     EXPECT_EQ(0x1ad92871U, r2.minimum_ttl());
 }
+
+TEST_F(DNSTest, BadLabelSize) {
+    constexpr uint8_t header[] = {
```
mfontanini, line 556: I think visual studio isn't liking this. Not sure if it's because of `constexpr`, this looks legal but the compiler may be a bit outdated.

src/dns.cpp
bwillcox: the lower 6 bits of this byte are the upper bits of the offset, so those bits can be non-zero. the check must for just the upper two bits of the byte

tests/src/dns_test.cpp
bwillcox: yeah seems like a subset of the compilers don't recognize constexpr yet; changed those to const...

src/dns.cpp
mfontanini: Fair enough, I misread what the problem was.

mfontanini (approved):
Thanks for the fix!

PR #468, opened 2022-02-24
dns: parser reads into garbage on misreported packet size

bwillcox:
The unit test demonstrates the following issue. The DNS packet is valid, but the packet size is misreported to be larger than the size of the valid packet. Attempting to get RR's for the additional section causes the parser to read into garbage past the end of the packet. This results in a malformed packet exception.

There is no language in the spec that addresses this use case. However, the RR counts in the header provide sufficient data for the parser to know when it could stop parsing.

Added code to limit RR fetch to count specified in header.


mfontanini:
Nice one. I'm curious whether you found this in the wild and how it would actually come to be? e.g. what would the trailer padding be in a real packet. Regardless, this seems reasonable so merging!

PR #471, opened 2022-04-24
android workaround

lattice0:


lattice0:
Would this workaround be ok so we can get android working?

mfontanini:
Thanks!

PR #478, opened 2022-05-28
clang-tidy fixes

neheb:


neheb:
why is there still C++98 support?

PR #481, opened 2022-06-21
Remove deprecated std::iterator

ecatmur:
replace with direct typedefs

### DIFF
--- a/include/tins/sniffer.h
+++ b/include/tins/sniffer.h
@@ -451,11 +451,11 @@ HandlerProxy<T> make_sniffer_handler(T* ptr,
  */
 class SnifferIterator {
 public:
-    typedef std::forward_iterator_tag iterator_category;
-    typedef Packet value_type;
-    typedef std::ptrdiff_t difference_type;
-    typedef Packet* pointer;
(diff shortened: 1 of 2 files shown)


include/tins/sniffer.h
```diff
@@ -449,8 +449,14 @@ HandlerProxy<T> make_sniffer_handler(T* ptr,
 /**
  * \brief Iterates over packets sniffed by a BaseSniffer.
  */
-class SnifferIterator : public std::iterator<std::forward_iterator_tag, Packet> {
+class SnifferIterator {
 public:
+    typedef std::forward_iterator_tag iterator_category;
```
neheb, line 454: Maybe replace typedef with using

ecatmur: OK, done

mfontanini: Unfortunately the default build mode still supports C++03 so we can't use `using` here and have to stick to typedef. It's probably about time to drop support for C++03 but we can't just yet.

mfontanini:
Thanks and sorry for taking a year to reply!

Issue #484, opened 2022-08-18
Incorrect IP Address range calculation when using /0 prefix


mbcdev:
It looks like IP address range calculations are not working correctly when using a /0 prefix.

Test Results (using 64-Bit Windows):

```
auto host_ip = Tins::IPv4Address("10.2.34.10");
auto zero_ip = Tins::IPv4Address("0.0.0.0");
auto zero_mask = Tins::IPv4Address("0.0.0.0");

auto range1 = Tins::IPv4Range::from_mask(zero_ip, zero_mask);
auto is_in_range1 = range1.contains(host_ip); // true (correct!)

auto range2 = Tins::IPv4Address(zero_ip) / 0;
auto is_in_range2 = range2.contains(host_ip); // false (wrong, should be true!)

auto range3 = Tins::IPv4Range::from_mask(zero_ip, Tins::IPv4Address::from_prefix_length(0));
auto is_in_range3 = range3.contains(host_ip); // false (wrong, should be true as well!)

auto mask0 = Tins::IPv4Address::from_prefix_length(0);
auto mask32 = Tins::IPv4Address::from_prefix_length(32);
auto is_equal = (mask0 == mask32); // true (wrong, should be false!)
```

mbcdev:
Reason:
According to the C/C++ Standard, for shift operations, the behavior is undefined if the right operand is equal to the width of the promoted left operand.
On a 64-bit Windows machine, this causes IP addresses 0.0.0.0 and 255.255.255.255 to have the same internal representation, leading to various issues when using a /0 prefix.

fixed in #486

mbcdev:
merged, closing issue

Issue #380, opened 2019-12-08
Missing wpcap.dll


KainWhite:
I have started up with this code:

yeahitsjan:
Don't know if this is still relevant for you. But instead also define ``WIN32`` in your file. ``pcap.h`` shows this:

```
#if defined(WIN32)
  #include <pcap-stdinc.h>
#elif defined(MSDOS)
  #include <sys/types.h>
  #include <sys/socket.h>  /* u_int, u_char etc. */
#else /* UN*X */
  #include <sys/types.h>
  #include <sys/time.h>
#endif /* WIN32/MSDOS/UN*X */
```

Should work. Got the same error and fixed it with that.

rectified95:
@yeahitsjan Do you mean you were able to fix the missing DLL error by defining WIN32 in the source file?

mfontanini:
I don't use Windows so it's hard for me to reproduce this. This was working fine last I tried it there, but I'm unsure of what version of Windows I used at the time.

rectified95:
@mfontanini Here are a few issues on Windows 11 I wanted to bring to your attention:
- Libtins website lists WinPCAP as a dependency on Windows: that project hasn't worked on Windows for a while, and itself instructs users to use NPCAP instead, which has a compatible API - https://npcap.com/
- The build artifacts expire after a month, so until yesterday it was impossible to access them and avoid building from source. It'd be awesome if you could extend the retention time or host the binaries somewhere else.

I'm gonna try the pre-built version and update the comment, but letting you know that Libtins seems broken on Windows when building from source.
The missing wpcap.dll error is odd given that it occurs after static compilation.

rectified95:
I was able to figure it out by reading the NPCAP docs - they do ship wpcap.dll.
I only downloaded the NPCAP SDK - analogous to libtins saying to get the WinPCAP SDK - but didn't install NPCAP itself - doing that solved this issue.
I think it can be closed. FYI @mfontanini @KainWhite

PR #486, opened 2022-08-26
fix incorrect IP address range calculation when using /0 prefix (#484)

mbcdev:
According to the C/C++ Standard, for shift operations, the behavior is undefined if the right operand is equal to the width of the promoted left operand.
On a 64-bit Windows machine, this causes IP addresses 0.0.0.0 and 255.255.255.255 to have the same internal representation, leading to various issues when using a /0 prefix.


rectified95:
@mbcdev Sorry about this totally unrelated question but there seems to be very little traffic on this repo otherwise - I assume you're able to use libtins in an app.
I'm getting an error about missing wpcap.dll (like this one: https://github.com/mfontanini/libtins/issues/380)
and wonder how to mitigate it. 
I'm using npcap.

mfontanini (approved):

mfontanini:
Thanks for the fix! @rectified95 there's not much traffic but this library is working fine and you can use it in applications. There's some companies using it in their products :)

PR #488, opened 2022-09-07
Qualify calls to std::move (clang 15)

ecatmur:
unqualified call to 'std::move' [-Wunqualified-std-cast-call]


ecatmur:
see: https://reviews.llvm.org/D119670

mfontanini:
:pray:

PR #494, opened 2022-12-13
validate high order two bits of first dns label octet

bwillcox:


src/dns.cpp
```diff
@@ -86,14 +86,17 @@ void DNS::skip_to_dname_end(InputMemoryStream& stream) const {
             break;
         }
         else {
-            if ((value & 0xc0)) {
+            if ((value & 0xc0) == 0xc0) {
```
mfontanini, line 89: Can we make both checks against the masked value so that it's clear you're expecting _that_ to be either X or Y? Very nitpicky but checking the same 2 bits in 2 different ways is harder to reason about.

```suggestion
            const uint8_t offset_discriminator = value & 0xc0;
            if (offset_discriminator == 0xc0) {
                // This is an offset label, skip the second byte and we're done
                stream.skip(1);
                break;
            }
            else if (offset_discriminator == 0) {
```

src/dns.cpp
```diff
@@ -86,14 +86,23 @@ void DNS::skip_to_dname_end(InputMemoryStream& stream) const {
...
                 stream.skip(1);
                 break;
             }
-            else {
+            else if (offset_discriminator == 0) {
+                // This is an offset label, skip the second byte and we're done
+                stream.skip(1);
+                break;
+            }
+            else if ((value & 0x3f) == value) {
```
mfontanini, line 100: Uh I think the diff was bad, this is duplicated now. Sorry!

mfontanini:
Thanks!

PR #496, opened 2023-01-24
Fix build with GCC13

Vascom:
Due to changes in GCC13 need fix include.
https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes

Else this error happen
```
/builddir/build/BUILD/libtins-4.4/src/../include/tins/ip_address.h:265:31: error: 'uint32_t' is not a member of 'std'; did you mean 'wint_t'?
  265 |         return std::hash<std::uint32_t>()(addr);
      |                               ^~~~~~~~
```

### DIFF
--- a/include/tins/ip_address.h
+++ b/include/tins/ip_address.h
@@ -33,7 +33,7 @@
 #include <string>
 #include <iosfwd>
 #include <functional>
-#include <stdint.h>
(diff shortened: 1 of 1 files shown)


mfontanini:
Nice, thanks!

PR #499, opened 2023-03-05
Add FileSniffer constructor with FILE pointer as pcap source

peckto:
This PR adds two overloads for the `FileSniffer` class:

```c++
FileSniffer(FILE *fp, const SnifferConfiguration& configuration);
FileSniffer(FILE *fp, const std::string& filter = "");
```

The possibility of opening a pcap by providing a FILE pointer adds more flexibility, e.g. for working with pipes.


mfontanini:
Thanks!

PR #500, opened 2023-04-20
Ignore IPv6 packets with payload after one with no Next Header

jamestiotio:
IPv6 data packets with payload or padded bytes received after one with no Next Header were not being parsed correctly, resulting in NULL PDU.

This commit fixes the IPv6 parser to be compliant with RFC 2460 Section 4.7 by adding a check in the IPv6 constructor to ignore the subsequent packets if an IPv6 packet contains no Next Header.

Signed-off-by: James Raphael Tiovalen <jamestiotio@gmail.com>


mfontanini:
Thanks!

PR #501, opened 2023-04-30
Add VXLAN support

### LINKED ISSUE #16: VXLan support
Hi,

First of all, thanks for developing Libtins! I would like to ask an enhancement:
- Support for VXLAN encapsulation :)
- A tutorial on how to create a new PDU type (so that I can write the encap myself)

jamestiotio:
This patch adds a new PDU class to support VXLAN. Several VXLAN-related tests have also been added.

This PR closes #16.

Signed-off-by: James Raphael Tiovalen <jamestiotio@gmail.com>

### DIFF
--- a/include/tins/vxlan.h
+++ b/include/tins/vxlan.h
@@ -1,0 +1,98 @@
+#ifndef TINS_VXLAN_H
+#define TINS_VXLAN_H
+
+#include <tins/pdu.h>
+#include <tins/small_uint.h>
+
+namespace Tins {
+
+/** 
+ * \class VXLAN
+ * \brief Represents a VXLAN PDU.
+ *
+ * This class represents a VXLAN PDU.
+ *
+ * \sa RawPDU
+ */
+class TINS_API VXLAN : public PDU {
+public:
+    /**
+     * \brief This PDU's flag.
+     */
+    static const PDU::PDUType pdu_flag = PDU::VXLAN;
+
+    /**
+     * \brief Constructs a VXLAN PDU.
+     *
+     * \param vni VXLAN Network Identifier.
+     */
+    VXLAN(const small_uint<24> vni = 0);
+
+    /**
+     * \brief Constructs a VXLAN object from a buffer and adds
+     * the Ethernet II PDU.
+     *
+     * \param buffer The buffer from which this PDU will be constructed.
+     * \param total_sz The total size of the buffer.
+     */
+    VXLAN(const uint8_t* buffer, uint32_t total_sz);
+
+    /**
+     * \brief Getter for the flags.
+     */
+    uint8_t get_flags() const { return Endian::be_to_host(header_.flags) >> 24; }
+
+    /**
+     * \brief Getter for the VNI.
+     */
+    small_uint<24> get_vni() const { return Endian::be_to_host(header_.vni) >> 8; }
+
+    /**
+     * \brief Setter for the flags.
+     * \param new_flags The new flags.
+     */
+    void set_flags(uint8_t new_flags) { header_.flags = Endian::host_to_be(new_flags << 24); }
+
+    /**
+     * \brief Setter for the VNI.
+     * \param new_vni The new VNI.
+     */
+    void set_vni(small_uint<24> new_vni) { header_.vni = Endian::host_to_be(new_vni << 8); }
+
+    /**
+     * \brief Returns the VXLAN frame's header length.
+     * 
+     * This method overrides PDU::header_size. This size includes the
+     * payload and options size.
+     *
+     * \return An uint32_t with the header's size.
+     * \sa PDU::header_size
+     */
+    uint32_t header_size() const { return sizeof(header_); }
+
+    /**
+     * \brief Getter for the PDU's type.
+     * \sa PDU::pdu_type
+     */
+    PDUType pdu_type() const { return pdu_flag; }
+
+    /**
+     * \sa PDU::clone
+     */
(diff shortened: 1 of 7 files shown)


src/vxlan.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+VXLAN::VXLAN(const uint32_t vni) {
+    header_.flags = 0x08000000;
+    header_.vni = vni << 8;
+}
+
+VXLAN::VXLAN(const uint8_t* buffer, uint32_t total_sz) {
+    InputMemoryStream stream(buffer, total_sz);
+    stream.read(header_);
+    header_.flags = ntohl(header_.flags);
```
mfontanini, line 23: Typically the pattern all other PDUs follow is:
* The constructor from a buffer just copies things as-is. This means it will preserve the protocol's endianness. 
* `write_serialization` copies it as-is.
* `get_XYZ` and `set_XYZ` handle the conversion by using `Endian::host_to_be` and `Endian::be_to_host` (those handle the type as well so you don't have to call one specific to e.g. 32 bit ints).

I'm not saying the way you're doing it is wrong, but I'd prefer if all PDUs did it consistently.

include/tins/vxlan.h
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * \param buffer The buffer from which this PDU will be constructed.
+     * \param total_sz The total size of the buffer.
+     */
+    VXLAN(const uint8_t* buffer, uint32_t total_sz);
+
+    /**
+     * \brief Getter for the flags.
+     */
+    uint32_t get_flags() const { return header_.flags >> 24; }
```
mfontanini, line 42: I think this should be a `uint8_t`?

include/tins/vxlan.h
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * \brief Getter for the VNI.
+     */
+    uint32_t get_vni() const { return header_.vni >> 8; }
```
mfontanini, line 47: Same here, this one should be a `small_uint<24>`.

include/tins/vxlan.h
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * \brief Returns the VXLAN frame's header length.
```
mfontanini, line 50: Could you add a setter for each of the 2 fields using the types mentioned above?

tests/src/vxlan_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+const uint32_t VXLANTest::vni = 0xffffff;
+const uint32_t VXLANTest::flags = 8;
+const IP::address_type VXLANTest::dst_ip = IP::address_type{"2.2.2.2"};
+const IP::address_type VXLANTest::src_ip = IP::address_type{"1.1.1.1"};
+const EthernetII::address_type VXLANTest::dst_addr = EthernetII::address_type{"aa:bb:cc:dd:ee:ff"};
+const EthernetII::address_type VXLANTest::src_addr = EthernetII::address_type{"8a:8b:8c:8d:8e:8f"};
+
+TEST_F(VXLANTest, Flags) {
+    auto const vxlan = VXLAN{};
+    EXPECT_TRUE(vxlan.get_flags() == flags);
```
mfontanini, line 46: nit: all of these `EXPECT_TRUE(x == y)` can be written as `EXPECT_EQ(x, y)` which will save some keystrokes and make the error message in case of failure cleaner.

mfontanini (commented):
A couple of minor comments. Thanks!

include/tins/vxlan.h
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * \brief Getter for the VNI.
+     */
+    small_uint<24> get_vni() const { return Endian::be_to_host(header_.vni) >> 8; }
+
+    /**
+     * \brief Setter for the flags.
+     * \param new_flags The new flags.
+     */
+    void set_flags(uint8_t new_flags) { header_.flags = Endian::host_to_be(new_flags); }
```
mfontanini, line 54: Shouldn't there be a shift here?

src/vxlan.cpp
jamestiotio: Sure thing. I have modified and moved things around to follow the standard PDU implementation. Do let me know if any further changes are required.

include/tins/vxlan.h
jamestiotio: Yup, you are right. I have changed it in the latest commit.

include/tins/vxlan.h
jamestiotio: Got it.

include/tins/vxlan.h
jamestiotio: I have added a setter for each of the 2 fields in the latest commit.

tests/src/vxlan_test.cpp
jamestiotio: Sure, I have changed these to `EXPECT_EQ(x, y)` in the latest commit.

include/tins/vxlan.h
jamestiotio: Yup. Apologies for the multiple forced pushes. I was a little confused about the order between `host_to_be`/`be_to_host` and the bit shifting but I managed to figure it out. The latest commit should be working properly and all `VXLANTest` tests are passing.

jamestiotio:
Hi @mfontanini, I have resolved your concerns with the latest commit. Do let me know if you have any other suggestions or comments. Thank you!

mfontanini:
Awesome, thanks!

PR #526, opened 2024-02-20
Resolves #525 Android 24+ support

### LINKED ISSUE #525: Seeing "android ifaddr not supported" regardless of Android version
Android added support for `getifaddrs` and `freeifaddrs` in Android API Level 24: https://android.googlesource.com/platform/bionic/+/master/libc/include/ifaddrs.h

In `routing_utils.cpp` the `network_interfaces()` function and in the `NetworkInterface` class the `info()` function has preprocessor directives that prevent Android systems with API Level >= 24 from using these functions even though they have support for `getifaddrs` and `freeifaddrs`.

I'll open up a PR for this issue that checks that the Android API level is less than 24 before throwing this exception

camm73:
Update to allow calls to `getifaddrs()` and `freeifaddrs` on Android systems with API Level 24+

Support in Android: https://android.googlesource.com/platform/bionic/+/master/libc/include/ifaddrs.h

### DIFF
--- a/src/network_interface.cpp
+++ b/src/network_interface.cpp
@@ -331,7 +331,7 @@ NetworkInterface::Info NetworkInterface::info() const {
     }
 
     #else // _WIN32
-    #ifndef ANDROID 
+    #if !defined(ANDROID) || (defined(__ANDROID_API__) && __ANDROID_API__ >= 24)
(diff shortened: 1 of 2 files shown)


mfontanini:
Thanks!

PR #527, opened 2024-03-05
Replace remaining `rfind_pdu` calls to `find_pdu` in `src` directory

jamestiotio:
This commit replaces the remaining `rfind_pdu` calls to `find_pdu` in the `src` directory. Any existing `rfind_pdu` calls in the `examples` and `tests` directories are unmodified.

The main motivation is that conditional statements are generally more performant than exception handling. Since `rfind_pdu` calls `find_pdu` internally anyway, this eliminates some overhead as well.


mfontanini:
Thanks!

PR #528, opened 2024-03-10
Update VXLANTest destination port number

jamestiotio:
Based on RFC 7348, the default IANA-assigned destination UDP port number for VXLAN is 4789, not the source port number. Update the value in the VXLANTest class to align with the RFC.

The VXLAN implementation still allows for the destination port number to be configurable to allow interoperability (as specified by the RFC), such as with the Linux kernel's default value of 8472.


mfontanini:
Thanks!

PR #533, opened 2024-06-13
Add BFD support

jamestiotio:
Add support for the Bidirectional Forwarding Detection (BFD) protocol as defined in RFC 5880. Some tests have also been added to ensure that the BFD PDU class functionalities are working as expected.


jamestiotio:
Hmm interesting, I didn't see the packed field bind compile error when testing locally. I will investigate this later.

